PR #27, opened 2020-05-06
Feature/permission set detail

annashamray:
Fixes https://github.com/GemeenteUtrecht/ZGW/issues/444

### DIFF
--- a/src/zac/accounts/templates/accounts/permissionset_detail.html
+++ b/src/zac/accounts/templates/accounts/permissionset_detail.html
@@ -1,0 +1,59 @@
+{% extends "master.html" %}
+{% load i18n sniplates %}
+
+{% block nav-main-id %}
+<article class="object-id">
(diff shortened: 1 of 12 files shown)


src/zac/accounts/templates/accounts/permissionset_detail.html
```diff
@@ -0,0 +1,59 @@
...
+{% block nav-main-id %}
+<article class="object-id">
+    <h1 class="object-id__identifier" title="Identificatie">
+        {{ permission_set.name }}
+    </h1>
+</article>
+{% endblock %}
+
+{% block content %}
+{% load_widgets kv='sniplates/key-value.html' %}
```
sergei-maertens, line 13: I believe this is unused?

annashamray: yes, it's what left after some experiments with formatting :)
I'll remove it in the next commits

PR #34, opened 2020-05-29
upload documents for advisors

annashamray:
Fixes: https://github.com/GemeenteUtrecht/ZGW/issues/462

### DIFF
--- a/src/zac/advices/forms.py
+++ b/src/zac/advices/forms.py
@@ -3,7 +3,7 @@ from django import forms
 from django_camunda.api import get_process_instance_variable
 
 from zac.core.forms import TaskFormMixin, _repr
-from zac.core.services import get_documenten, get_zaak
+from zac.core.services import get_documenten, get_zaak, update_document
 
 from .constants import AdviceObjectTypes
 from .models import Advice
@@ -38,12 +38,15 @@ class AdviceForm(TaskFormMixin, forms.ModelForm):
 class UploadDocumentForm(forms.Form):
     url = forms.CharField(widget=forms.HiddenInput())
     titel = forms.CharField()
-    upload = forms.FileField(widget=forms.FileInput(), label="Upload new version")
+    upload = forms.FileField(
+        widget=forms.FileInput(), label="Upload new version", required=False
+    )
 
 
 class UploadDocumentBaseFormSet(forms.BaseFormSet):
-    def __init__(self, task, *args, **kwargs):
+    def __init__(self, task, user, *args, **kwargs):
         self.task = task
+        self.user = user
 
         if "initial" in kwargs:
             super().__init__(**kwargs)
@@ -59,6 +62,24 @@ class UploadDocumentBaseFormSet(forms.BaseFormSet):
         kwargs["initial"] = initial
 
         super().__init__(*args, **kwargs)
+
+    def on_submission(self):
+        assert self.is_valid(), "FormSet must be validated first"
+        updated_eios = []
+        for form in self.forms:
+            if form.has_changed() and form.cleaned_data["upload"]:
+                data = {
+                    "titel": form.cleaned_data["titel"],
+                    "auter": self.user.username,
+                }
+                # todo add auter from request!
+                document = update_document(
+                    url=form.cleaned_data["url"],
+                    data=data,
+                    file=form.cleaned_data["upload"],
+                )
+                updated_eios.append(document)
+        print("updated eios = ", updated_eios)
 
 
 UploadDocumentFormset = forms.formset_factory(
--- a/src/zac/advices/forms.py
+++ b/src/zac/advices/forms.py
@@ -36,8 +36,31 @@ class AdviceForm(TaskFormMixin, forms.ModelForm):
 
 
 class UploadDocumentForm(forms.Form):
-    doc_id = forms.CharField()
-    new_version = forms.FileField()
+    url = forms.CharField(widget=forms.HiddenInput())
+    titel = forms.CharField()
+    upload = forms.FileField(widget=forms.FileInput(), label="Upload new version")
 
 
-UploadDocumentFormset = forms.formset_factory(UploadDocumentForm, extra=1)
+class UploadDocumentBaseFormSet(forms.BaseFormSet):
+    def __init__(self, task, *args, **kwargs):
+        self.task = task
+
+        if "initial" in kwargs:
+            super().__init__(**kwargs)
+
+        # retrieve process instance variables
+        zaak_url = get_process_instance_variable(
+            self.task.process_instance_id, "zaakUrl"
+        )
+        zaak = get_zaak(zaak_url=zaak_url)
+        documenten, _ = get_documenten(zaak)
+
+        initial = [{"url": doc.url, "titel": doc.titel} for doc in documenten]
+        kwargs["initial"] = initial
+
+        super().__init__(*args, **kwargs)
+
+
+UploadDocumentFormset = forms.formset_factory(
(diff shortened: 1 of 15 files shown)


src/zac/advices/forms.py
```diff
@@ -32,3 +34,60 @@ def on_submission(self):
...
+    )
+
+
+class UploadDocumentBaseFormSet(forms.BaseFormSet):
+    def __init__(self, task, user, *args, **kwargs):
+        self.task = task
+        self.user = user
+
+        if "initial" in kwargs:
+            super().__init__(**kwargs)
```
sergei-maertens, line 53: hrm, so if `initial` is provided, that'll run the `super().__init__` twice - I don't think that's a good idea.

I'd refactor this into figuring out the `initial` and using `kwargs.setdefault("initial", initial)`

src/zac/advices/forms.py
```diff
@@ -32,3 +34,60 @@ def on_submission(self):
     def get_process_variables(self):
         # does not set any variables
         return {}
+
+
+class UploadDocumentForm(forms.Form):
+    url = forms.CharField(widget=forms.HiddenInput())
+    titel = forms.CharField()
```
sergei-maertens, line 41: I'd leave the title alone, only the content matters afaik

src/zac/advices/forms.py
```diff
@@ -32,3 +34,60 @@ def on_submission(self):
...
+    def on_submission(self):
+        assert self.is_valid(), "FormSet must be validated first"
+
+        changed = False
+        for form in self.forms:
+            if form.has_changed() and form.cleaned_data["upload"]:
+                changed = True
+                data = {
+                    "titel": form.cleaned_data["titel"],
+                    "auter": self.user.username,
```
sergei-maertens, line 78: typo
```suggestion
                    "auteur": self.user.username,
```

src/zac/advices/forms.py
```diff
@@ -32,3 +34,60 @@ def on_submission(self):
     def get_process_variables(self):
         # does not set any variables
         return {}
+
+
+class UploadDocumentForm(forms.Form):
+    url = forms.CharField(widget=forms.HiddenInput())
```
sergei-maertens, line 40: missing validation that the URL does indeed belong to the zaak from the task. Now someone can edit the hidden input to update a document they don't have permissions for, for example.

src/zac/advices/forms.py
```diff
@@ -32,3 +34,60 @@ def on_submission(self):
...
+
+        changed = False
+        for form in self.forms:
+            if form.has_changed() and form.cleaned_data["upload"]:
+                changed = True
+                data = {
+                    "titel": form.cleaned_data["titel"],
+                    "auter": self.user.username,
+                }
+                document = update_document(
```
sergei-maertens, line 80: assigned variable is not used

src/zac/core/cache.py
```diff
@@ -50,3 +52,22 @@ def invalidate_zaak_list_cache(client: Client, zaak: Zaak):
...
+    from .services import get_zaak_informatieobjecten
+
+    zaak_informatieobjecten = get_zaak_informatieobjecten(zaak)
+    zios = [zio["informatieobject"] for zio in zaak_informatieobjecten]
+
+    # construct cache keys
+    permutations = itertools.permutations(zios)
+    for permutation in permutations:
+        key = "zios:{}".format(",".join(permutation))
+        key = hashlib.md5(key.encode("ascii")).hexdigest()
```
sergei-maertens, line 72: I'd rather move the cache-key calculation to a shared utility function instead of duplicating the implementation

src/zac/core/services.py
```diff
@@ -689,3 +694,37 @@ async def fetch_documents(zios: list):
         responses = await asyncio.gather(*tasks)
 
     return responses
+
+
+def update_document(url, file, data):
```
sergei-maertens, line 699: I'd like to see some type hints here

src/zac/core/templates/core/zaak_task.html
```diff
@@ -49,6 +49,24 @@
...
 
+        {% if formset %}
+            <section class="content-panel formset">
+            {{ formset.management_form }}
+            {{ formset.non_form_errors }}
+
+            {% for inline_form in formset %}
+                {{  inline_form.non_field_errors }}
+                <div class="form__field-group formset__group">
+                {% for field in inline_form %}
```
sergei-maertens, line 60: :lipstick: can you make sure there's no border around the file input?

I'd also prefer a table layout for the formset - currently for every form in the formset, the labels are repeated. I'd rather see something like:

| Source document | Upload edited document |
|---|---|
| [file.docx](#) | `<input type="file">` |
| [important.pdf](#) | `<input type="file">` |

src/zac/core/views/processes.py
```diff
@@ -117,7 +117,34 @@ def form_valid(self, form: ClaimTaskForm):
...
+
+    def get_formset(self, **kwargs):
+        formset_class = self.get_formset_class()
+        if not formset_class:
+            return
+
+        # FIXME how to feed initial data into formset ???
+        # kwargs = {"initial": formset_class.get_initial()}
+
+        if self.request.method in ("POST", "PUT"):
```
sergei-maertens, line 133: not sure why PUT is in here
```suggestion
        if self.request.method == "POST":
```

src/zac/core/views/processes.py
```diff
@@ -117,7 +117,34 @@ def form_valid(self, form: ClaimTaskForm):
...
+            kwargs.update(
+                {"data": self.request.POST.copy(), "files": self.request.FILES}
+            )
+        formset = formset_class(**kwargs, task=self._get_task(), user=self.request.user)
+        return formset
+
+    def get_context_data(self, **kwargs):
+        if "formset" not in kwargs:
+            kwargs["formset"] = self.get_formset(**kwargs)
```
sergei-maertens, line 142: not sure if putting _all_ the context data as formset kwargs is a good idea. Maybe better to define `get_formset_kwargs` and use that, and in the view you can then pass explicitly what's needed.

src/zac/core/views/processes.py
```diff
@@ -181,6 +208,17 @@ def get_success_url(self):
         )
 
     def form_valid(self, form):
+        context = self.get_context_data()
+        formset = context["formset"]
```
sergei-maertens, line 212: I'd rather see `self.get_formset()` used here instead

src/zac/core/views/processes.py
annashamray: `get_formset_kwargs` method is added

annashamray:
> There is something odd going on after I update a document as well - in the main case, the URL bit that should contain the `identificatie` shows `None`, and when I try to start another advice case, relating the document fails: https://bptl.utrechtproeftuin.nl/tasks/5341/ I'm 99% sure it's a bug in drc-cmis, so maybe can you include `identitificatie` as part of the PATCH body?

@sergei-maertens I added `identificatie` field to the formset, but it doesn't seep to help. The response from DRC has `identificatie=None`

sergei-maertens:
> > There is something odd going on after I update a document as well - in the main case, the URL bit that should contain the `identificatie` shows `None`, and when I try to start another advice case, relating the document fails: https://bptl.utrechtproeftuin.nl/tasks/5341/ I'm 99% sure it's a bug in drc-cmis, so maybe can you include `identitificatie` as part of the PATCH body?
> 
> @sergei-maertens I added `identificatie` field to the formset, but it doesn't seep to help. The response from DRC has `identificatie=None`

right, I was afraid this would happen. Guess I'm patching drc-cmis again... :(

PR #40, opened 2020-06-30
Adding approval

SilviaAmAm:


src/zac/advices/templates/advices/includes/approval_table.html
```diff
@@ -0,0 +1,27 @@
...
+    <tbody>
+        {% for approval in approvals %}
+            <tr>
+                <td>
+                    {% if approval.approved %} {% trans "Approved" %} {% else %} {% trans "Not approved" %} {% endif %}
+                </td>
+                <td> {% firstof approval.author.get_full_name approval.author.username %}</td>
+                <td>
+                    <time title="{{ approval.created.isoformat }}">
+                        {{ approval.created|timesince }} {% trans "ago" %}
```
sergei-maertens, line 21: `{% blocktrans %}` is better since it allows you to re-arrange the order of sentence parts

src/zac/core/forms.py
```diff
@@ -287,3 +289,19 @@ def get_process_variables(self) -> Dict[str, List[str]]:
             "documenten": self.cleaned_data["documenten"],
             "kownslReviewRequestId": self.cleaned_data["review_request"],
         }
+
+
+class ConfigureApprovalRequestForm(ConfigureAdviceRequestForm):
```
sergei-maertens, line 294: The `on_submission` is now creating a review request of the wrong type.

Probably there should be a base class, inherited for both advice/approval and you could leverage a class property to set the type of review request to keep the `on_submission` and `get_process_variables` generic.

src/zac/advices/templates/advices/includes/approval_table.html
```diff
@@ -0,0 +1,31 @@
...
+                        {% blocktrans %}Approved{% endblocktrans %}
+                    {% else %}
+                        {% blocktrans %}Not approved{% endblocktrans %}
+                    {% endif %}
+                </td>
+                <td> {% firstof approval.author.get_full_name approval.author.username %}</td>
+                <td>
+                    <time title="{{ approval.created.isoformat }}">
+                        {{ approval.created|timesince }} {% blocktrans %}ago{% endblocktrans %}
+                    </time>
```
sergei-maertens, line 26: Please review the documentation for [`blocktrans`](https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#std:templatetag-blocktrans) - 

```suggestion
                    <time title="{{ approval.created.isoformat }}">
                        {% blocktrans with created_since=approval.created|timesince trimmed %}
                            {{ created_since }} ago
                        {% endblocktrans %}
                    </time>
```

The importance here is that the order of `5 days ago` is not the same in every language - other language could have `foo {{ created_since }}` in the translated form. That's why you need to put the entire construct in the `{% blocktrans %}` tags.

src/zac/core/forms.py
```diff
@@ -287,3 +286,34 @@ def get_process_variables(self) -> Dict[str, List[str]]:
...
+            review_request = create_review_request(
+                self.zaak_url,
+                review_type=self._review_type,
+                num_assigned_users=len(self.cleaned_data["users"]),
+            )
+            self.cleaned_data["review_request"] = str(review_request.id)
+        else:
+            raise NotImplementedError(
+                f"Variable _review_type needs to be defined in {self.__name__}"
+            )
```
sergei-maertens, line 301: This defensive style goes a bit against the zen of python (better to ask forgiveness than permission), so I'd replace that with:

```suggestion
        assert self._review_type, "Subclasses must define a '_review_type'"
        review_request = create_review_request(
            self.zaak_url,
            review_type=self._review_type,
            num_assigned_users=len(self.cleaned_data["users"]),
        )
        self.cleaned_data["review_request"] = str(review_request.id)
```

src/zac/core/forms.py
```diff
@@ -287,3 +286,34 @@ def get_process_variables(self) -> Dict[str, List[str]]:
...
+
+
+class ConfigureAdviceRequestForm(ConfigureReviewRequestForm):
+    """
+    Child class of ConfigureReviewRequestForm which specifies that the type of review requested
```
sergei-maertens, line 306: Docstrings should follow the format of:

```
Do some action to achieve something.

More elaborate explanation of what's happening, attribute documentation... etc.
```

Suggested docstring:
```suggestion
    Create an "advice" type of review request on valid submission.
```

src/zac/core/forms.py
```diff
@@ -287,3 +286,34 @@ def get_process_variables(self) -> Dict[str, List[str]]:
...
+    """
+
+    _review_type = "advice"
+
+
+class ConfigureApprovalRequestForm(ConfigureReviewRequestForm):
+    """
+    Child class of ConfigureReviewRequestForm which specifies that the type of review requested
+    is an advice request.
+    """
```
sergei-maertens, line 317: Suggested docstring:

```suggestion
    """
    Create an "approval" type of review request on valid submission.
    """
```

Issue #41, opened 2020-07-09
search for betrokkene  and display initiator 


annashamray:
1. In "Zoeken" enable "BRP" part:
    * add `BSN` search field
    * search through links to BRP in `betrokkene` field and through `betrokkeneIdentificatie.inpBsn` if `betrokkene` is missing

2. In ZaakDetail page in "BETROKKENEN" section:
    * display zaak initiator with their BSN, first name and last name. Use data from BRP request or from `betrokkeneIdentificatie` part in the Zaak

- [x] Search for BSN
- [x] Display BRP data

PR #43, opened 2020-07-14
Feature/display initiator

annashamray:
Issue - https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/41 (display initiator step)

### DIFF
--- a/src/zac/contrib/brp/api.py
+++ b/src/zac/contrib/brp/api.py
@@ -24,7 +24,6 @@ class HalClient(ZGWClient):
         headers["Accept"] = "application/hal+json"
         headers["Content-Type"] = "application/hal+json"
 
-        print(kwargs)
         return result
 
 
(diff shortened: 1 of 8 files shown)


src/zac/contrib/brp/api.py
```diff
@@ -0,0 +1,55 @@
...
+class HalClient(ZGWClient):
+    def pre_request(self, method, url, **kwargs):
+        """
+        Add authorization header to requests for APIs without jwt.
+        """
+        result = super().pre_request(method, url, **kwargs)
+
+        headers = kwargs.get("headers", {})
+        headers["Accept"] = "application/hal+json"
+        headers["Content-Type"] = "application/hal+json"
```
sergei-maertens, line 25: Don't think this is needed since we're not sending any data

annashamray: Yeah, I checked, only "Accept" header is needed. I can removed it later

PR #50, opened 2020-07-27
Feature/remove separate zaak for advices

annashamray:
Issue - https://github.com/GemeenteUtrecht/ZGW/issues/499


src/zac/core/views/zaken.py
```diff
@@ -85,31 +83,27 @@ def get_context_data(self, **kwargs):
...
+            _approvals = executor.map(
+                retrieve_approvals,
+                list(filter(lambda x: x.num_approvals, review_requests)),
             )
-            for review_zaak, review_request in zip(_review_zaken, review_requests):
-                review_request.review_zaak = review_zaak
+
+            related_zaken = _related_zaken.result()
+            advices = [result[0] for result in _advices]
+            approvals = [result[0] for result in _approvals]
```
sergei-maertens, line 102: This confuses me - the result variable `_advices` because of the `executor.map` should be a list of lists, each nested list being a collection of advices/approvals for a review request. Why are we only grabbing the first? Or is this part of the follow up PR for the display?

annashamray: you are right, I'll fix it in https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/51

PR #56, opened 2020-07-31
Process interface on zaak detail page

sergei-maertens:
- [x] Support tabs (a tab for every primary process)
- [x]  Support displaying the possible BPMN messages that can be sent
- [x] Display primary user tasks
    - [x] Display assignee status
    - [x] Implement claiming a task
    - [x] Implement 'executing' a task
- [x] Display secondary tasks (from subprocesses)
- [x] Re-implement sending BPMN messages


src/zac/camunda/processes.py
```diff
@@ -0,0 +1,79 @@
...
+from django_camunda.client import get_client
+from zgw_consumers.api_models.base import factory
+
+from zac.core.camunda import get_process_tasks
+
+from .data import ProcessInstance
+from .messages import get_messages
+
+
+def get_process_definition(definition_id: str) -> ProcessDefinition:
```
sergei-maertens, line 13: could use the `@cache` decorator - a definition is a particular version already and not expected to change, so caching for a day is suited.

src/zac/camunda/processes.py
```diff
@@ -0,0 +1,79 @@
...
+    client = get_client()
+
+    response = client.get(f"process-definition/{definition_id}")
+    return factory(ProcessDefinition, response)
+
+
+def add_subprocesses(process_instance, process_instances, client):
```
sergei-maertens, line 20: please add type hints

src/zac/camunda/processes.py
```diff
@@ -0,0 +1,79 @@
...
+
+    #  get process-instances for particular zaak
+    response = client.get("process-instance", {"variables": f"zaakUrl_eq_{zaak_url}"})
+
+    process_instances = {}
+    for data in response:
+        process_instance = factory(ProcessInstance, data)
+        process_instances[data["id"]] = process_instance
+
+    #  todo concurrency?
```
sergei-maertens, line 51: seems like a bad idea with threading and mutating a dict :)

src/zac/camunda/processes.py
```diff
@@ -0,0 +1,79 @@
...
+    # fill in all subpocesses into the dict
+    for id, process_instance in process_instances.copy().items():
+        add_subprocesses(process_instance, process_instances, client)
+
+    # add definitions and user tasks
+    definition_ids = {p.definition_id for p in process_instances.values()}
+    definitions = {
+        definition_id: get_process_definition(definition_id)
+        for definition_id in definition_ids
+    }
```
sergei-maertens, line 61: can be done more efficiently, since the [`/process-definition` endpoint](https://docs.camunda.org/manual/7.11/reference/rest/process-definition/get-query/) supports the querystring parameter `processDefinitionIdIn`

src/zac/camunda/processes.py
```diff
@@ -0,0 +1,79 @@
...
+    for id, process_instance in process_instances.items():
+        process_instance.definition = definitions[process_instance.definition_id]
+        if not process_instance.tasks:
+            process_instance.tasks = get_process_tasks(process_instance)
```
sergei-maertens, line 65: this is a candidate for `parallel()` context manager

src/zac/core/templates/core/zaak_detail.html
```diff
@@ -297,4 +295,7 @@ <h1 class="object-id__identifier" title="Identificatie">
 
 </article>
 
+{{ user.id|json_script:"userId" }}
+{{ user.username|json_script:"username" }}
```
sergei-maertens, line 299: TODO: serialize user, using the API defined serializer

src/zac/mock.json
```diff
@@ -0,0 +1,80 @@
+[
```
sergei-maertens, line 1: TODO: delete file & endpoint

src/zac/urls.py
```diff
@@ -29,3 +31,17 @@
...
+        import json
+        import os
+
+        path = os.path.join(os.path.dirname(__file__), "mock.json")
+        with open(path, "r") as mock_data:
+            data = json.loads(mock_data.read())
+        return JsonResponse(data, safe=False)
+
+
+urlpatterns += [path("api/mock", MockView.as_view(), name="mock-data")]
```
sergei-maertens, line 47:
```suggestion
```

src/zac/camunda/processes.py
annashamray: thanks! the request loop is replaced with a singular request

src/zac/camunda/processes.py
annashamray: this function is not used anymore, since it was replaced with `get_process_definition` with `processDefinitionIdIn` query param

src/zac/camunda/processes.py
annashamray: I couldn't make `parallel` work here. It seems that some DB lock happening in `get_process_tasks.get_client()`

src/zac/js/components/Tabs.js
```diff
@@ -0,0 +1,97 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+
+const Tab = ({ index, title, onClick, active=false }) => {
+    const target = `tab-content-${index}`;
+    const id = `tab-${index}`;
```
annashamray, line 7: this variable is not used here. Should I remove it?

src/zac/js/components/processes/ProcessInteraction.js
```diff
@@ -0,0 +1,206 @@
...
+        );
+        return num;
+    };
+    const numTasks = getNumTasks(processInstance);
+
+    return (
+        <React.Fragment>
+            {processInstance.title}
+            <span className="badge badge--spacing" title={`${numTasks} open taken`}>
+                { numTasks }
```
annashamray, line 120: Should we have a red circle around task number when there are 0 tasks?

src/zac/js/components/Tabs.js
sergei-maertens: hrm no, apparently I forgot to add it to the `li` element below. it's an accessibility thing

src/zac/js/components/processes/ProcessInteraction.js
sergei-maertens: I think we should just not display the badge if there are 0 tasks.

annashamray: agree! I'll make a change

PR #69, opened 2020-09-08
Added button to add documents on zaak-detail page

sergei-maertens:
* Adds permission to add documents (can be assigned via autorisatieprofielen)
* If permission is given, user sees a button to upload docs. API endpoint validates permissions.

### DIFF
--- a/src/zac/core/api/permissions.py
+++ b/src/zac/core/api/permissions.py
@@ -1,0 +1,35 @@
+import logging
+
+from rest_framework import permissions
+from rest_framework.request import Request
+from rest_framework.views import APIView
+from zds_client import ClientError
+
+from ..permissions import zaken_add_documents
+from ..services import get_zaak
+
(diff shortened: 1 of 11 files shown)


kelvincy (approved):
Future improvement: store csrf token in cookie

sergei-maertens:
> Future improvement: store csrf token in cookie

it already is in a cookie and used in traditional form submissions - django ensures this. However, since these pages don't contain forms, there's no guarantee that a CSRF token will be set (though we could probably use the `@ensure_csrf_cookie` decorator or something like that on the view)

But for AJAX calls, Django needs to extract this from the HTTP headers, and the token value can change with the user session. The CSRF cookie is also marked as httpOnly - so reading that from JS is not possible.

sergei-maertens:
Merging since Michelle wants to demo this tomorrow :grimacing:

Issue #65, opened 2020-08-21
Implement (temporary) access for a user to a specific Zaak


sergei-maertens:
Consider:

- a role "bezwaarjuristen", who have access to the zaaktype "Bezwaar behandelen"
- a zaak "zaak X" of zaaktype "Bezwaar behandelen", with a `relevanteAndereZaak` of zaaktype "Evenementenvergunning behandelen" (`aardRelatie` is `onderwerp`, but that's not a limiting factor)

A user "Joe" has the role "bezwaarjuristen" and is looking at "zaak X". The detail page shows that there is a related zaak.

When user "Joe" clicks the link to the bezwaarzaak, they should be presented with a permission denied screen, containing a "request access" form. The "request access" form should have a "remarks" field and submit button.

The access request(s) should be saved for a given zaak. A behandelaar of the zaak (for which access is requested) must be able to see that:
- there are (new) access requests
- who is requesting access, including the remarks

They must also be able to grant or reject the access.

Upon granting the access, user "Joe" should receive an e-mail that access was granted, and be able to access the zaak, including all documents, related objects... They should see a full picture.

Access is limited to that particular zaak, not to the entire zaaktype.

**Checklist**

- [x] Can see related zaken
- [x] Can click to related zaak and view permission denied page
- [x] Can submit access request from permission denied page
- [x] Zaakbehandelaar can see access requests
- [ ] Zaakbehandelaar is notified on access requests
- [x] Zaakbehandelaar can grant access
- [x] Zaakbehandelaar can reject access
- [x] Can access zaak detail page for which access was granted
- [x] The requester ("Joe") receives an email with access request result

PR #70, opened 2020-09-08
temporary access to zaken

annashamray:
fixes https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/65


src/zac/core/templates/core/zaak_access_requests.html
```diff
@@ -0,0 +1,54 @@
...
+
+
+{% block content %}
+<a href="{% url 'core:zaak-detail' bronorganisatie=zaak.bronorganisatie identificatie=zaak.identificatie %}" class="link link--backlink">
+    Terug naar de zaakgegevens
+</a>
+
+<header class="zaak-detail-page__header">
+    <h1 class="page-title" title="Zaaktype">
+        Access requests for "{{ zaak.zaaktype.omschrijving }}"-zaak
```
sergei-maertens, line 12: Mark this translatable please

src/zac/core/templates/core/zaak_access_requests.html
```diff
@@ -0,0 +1,54 @@
...
+            {{ zaak.zaaktype.omschrijving }}
+        </small>
+    </h1>
+</header>
+
+<form action="." method="post">
+    {% csrf_token %}
+
+    {{ formset.management_form }}
+    {{ formset.non_form_errors }}
```
sergei-maertens, line 24: Should be wrapped in appropriate widget :)

src/zac/core/templates/core/zaak_access_requests.html
```diff
@@ -0,0 +1,54 @@
...
+            <tr>
+                <th class="table__header">#</th>
+                <th class="table__header">User</th>
+                <th class="table__header">Comments</th>
+            </tr>
+        </thead>
+        <tbody>
+            {% for form in formset %}
+                <tr>
+                    {{  form.non_field_errors }}
```
sergei-maertens, line 37: this renders as an `<ul>` by default, which is invalid markup inside of a `<tr>` - you'll need to do some more error processing for formsets.

src/zac/core/tests/test_access_requests.py
```diff
@@ -0,0 +1,309 @@
...
+            response.html.find(class_="errorlist").text,
+            "You've already requested access for this zaak",
+        )
+
+        self.assertEqual(
+            AccessRequest.objects.exclude(id=previous_request.id).count(), 0
+        )
+
+
+@patch("zac.core.forms.send_mail")
```
sergei-maertens, line 182: not needed - django replaces the mail backend automatically and you can use `django.core.mail.outbox` to inspect the mails being sent.

src/zac/core/views/zaken.py
```diff
@@ -299,3 +307,73 @@ def get_object(self):
...
+
+
+class AccessRequestCreateView(LoginRequiredMixin, FormView):
+    form_class = AccessRequestCreateForm
+    template_name = "core/create_access_request.html"
+    #  todo add thanks page?
+    success_url = reverse_lazy("core:index")
+
+    def get_zaak(self):
+        return find_zaak(**self.request.resolver_match.kwargs)
```
sergei-maertens, line 319:
```suggestion
        return find_zaak(**self.kwargs)
```
Bit more consistent with the other views and this uses the top-level API instead of request resolver internals.

src/zac/core/views/zaken.py
```diff
@@ -299,3 +307,73 @@ def get_object(self):
...
+        zaak = find_zaak(**self.kwargs)
+        return zaak
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context.update({"zaak": self.get_zaak()})
+        return context
+
+    def formset_valid(self, formset):
+        result = self.request.POST["submit"]
```
sergei-maertens, line 371: input is not validated, so I can now POST any value for `submit`, which will be saved in the DB and then the admin will not be able to display the proper option. Edge case, but still - always validate all your inputs.

src/zac/templates/403.html
```diff
@@ -6,5 +6,11 @@
 
  <h1>Sorry, you don't have access to this page (403)</h1>
 
+{% if can_request_access %}
+    <p>
+        You can <a href="{% url 'core:access-request-create' bronorganisatie=zaak_kwargs.bronorganisatie identificatie=zaak_kwargs.identificatie %}">request access</a> for this page
```
sergei-maertens, line 11: Translations... :)

src/zac/templates/index.html
```diff
@@ -24,6 +24,12 @@ <h1 class="page-title" title="Zaaktype">{% trans "Werkvoorraad" %}</h1>
             class="tab__tab">
             <a href="#work-ad-hoc">Ad-hoc activiteiten ({{ adhoc_activities|length }})</a>
         </li>
+
+        <li role="tab" id="work-tab-access-requests"
+            aria-controls="work-access-requests"
+            class="tab__tab">
+            <a href="#work-access-requests">Access requests ({{ access_requests|length }})</a>
```
sergei-maertens, line 31: translations

src/zac/utils/views.py
```diff
@@ -23,3 +25,31 @@ def server_error(request, template_name=ERROR_500_TEMPLATE_NAME):
...
+        if template_name != ERROR_403_TEMPLATE_NAME:
+            # Reraise if it's a missing custom template.
+            raise
+        return http.HttpResponseForbidden(
+            "<h1>403 Forbidden</h1>", content_type="text/html"
+        )
+
+    context = {"exception": str(exception), "can_request_access": False}
+
+    if request.resolver_match.url_name == "zaak-detail" and request.user:
```
sergei-maertens, line 49:
```suggestion
    if request.resolver_match.url_name == "zaak-detail" and request.user.is_authenticated:
```

sergei-maertens (changes requested):
In general okay for happy flow, but note the permissions-related remarks - that could be a security regression!

In general, functionality wise, the following remarks:

* the current flow is that a Zaak must have "behandelaars" to be able to _request_ access. That's annoying for the requester, as it seems like it's their fault they can't request access. The requesting of access should always be possible, we can tackle who can grant it later if there's no suitable person.

* The 'zaak-detail' page/template has the side-nav to link to access requests, but there's no indication if there are any. If the count is non-zero, probably a good idea to display the count. Or, a toast/notification bar on top of the page saying that there are access requests pending would be a good visual indicator. This information (and the menu entry) should be limited to people who can grant access. (Probably a good idea to define a permission/rules predicate to check if a user can grant access or not, so that we can extend on this later).

src/zac/accounts/models.py
```diff
@@ -71,6 +74,11 @@ def get_short_name(self):
         "Returns the short name for the user."
         return self.first_name
 
+    def has_access_to_zaak(self, zaak: Zaak):
+        return self.initiated_requests.filter(
+            zaak=zaak.url, result=AccessRequestResult.approve
+        ).exists()
```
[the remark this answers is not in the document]
annashamray: this snippet is moved to `has_temporary_access` rule predicate

annashamray (commented):
specific permission `zaken_handle_access` is added.
But I have a couple of questions about navigation menu formatting:
* should other navigation items also show the counter (like activiteiten)? 
* should I implement the same logic for dashboard (werkvoorraad) and hide access request tab if user doesn't have `zaken_handle_access` permission?

src/zac/core/rules.py
```diff
@@ -83,4 +82,19 @@ def can_set_results(user: User, zaak: Optional[Zaak]):
     return _generic_zaakpermission(user, zaak, zaken_set_result)
 
 
+@rules.predicate
+def can_read_zaak_by_zaaktype(user: User, zaak: Optional[Zaak]):
+    if zaak is None:
+        return _has_permission_key(zaken_set_result.name, user)
+    return _generic_zaakpermission(user, zaak, zaken_inzien)
```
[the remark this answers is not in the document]
sergei-maertens: `Optional` is correct, but the `_has_permission_key(zaken_set_result.name, user)` is wrong, should be `_has_permission_key(zaken_inzien.name, user)` I think.

sergei-maertens:
> should other navigation items also show the counter (like activiteiten)?

No - we currently have that on the button on the top right, but we'll revisit probably at some point.

> should I implement the same logic for dashboard (werkvoorraad) and hide access request tab if user doesn't have zaken_handle_access permission?

Seems like a lot of work for something that doesn't really sit in the way if the list just stays empty.

PR #72, opened 2020-09-09
Initial display of zaakbesluiten

sergei-maertens:
Besluit can be added through Open Zaak admin for now.

The table-display does not nearly convey all the available information - most notably the documents that hold the actual "besluit". Given the timeline, this did not make it into MVP.


sergei-maertens:
Sufficient for Michelle for now

PR #77, opened 2020-09-15
:heavy_plus_sign: Add modal to retrieve extra BRP data

kelvincy:


src/zac/js/components/forms/Utils.js
```diff
@@ -1,21 +1,17 @@
...
-
-
-const Wrapper = ({ errors=[], children }) => {
+const Wrapper = ({ errors = [], children }) => {
     const hasErrors = errors && errors.length > 0;
-    let className = "input";
+    let className = 'input';
     if (hasErrors) {
-        className += " input--invalid";
+        className += ' input--invalid';
```
sergei-maertens, line 14: could benefit from `classnames` as well

src/zac/sass/components/_result.scss
```diff
@@ -0,0 +1,33 @@
...
+  &__value {
+    margin: 0;
+    display: inline-block;
+  }
+
+  &__row {
+    margin-left: 2rem;
+    margin-bottom: 2rem;
+  }
+}
```
sergei-maertens, line 33:
```suggestion
@import "../lib/bem";

.result {
  margin-bottom: 1rem;

  @include element("title") {
    margin: 0 0 3px 0;
    font-size: 1rem;
  }

  @include element("value") {
    margin: 0;
  }
}

.subresult {
  @include element("title") {
    font-weight: bold;
    display: inline-block;
    margin: 0 0.4rem 0 0;
    &:after {
      content: ':'
    }
  }

  @include element("value") {
    margin: 0;
    display: inline-block;
  }

  @include element("row") {
    margin-left: 2rem;
    margin-bottom: 2rem;
  }
}
```

this mixin ensures that the resulting CSS selector is `.block .block__element` rather than `.block__element` - the latter makes it possible to write markup with elements without a parent block. The strict CSS enforces proper markup/class names structure.

src/zac/sass/components/forms/_checkbox-select.scss
```diff
@@ -15,6 +15,59 @@
...
+    border: 0.1rem solid rgba(104, 119, 141, 0.3);
+    border-radius: 0.1rem;
+    outline: none;
+    transition: all 100ms;
+    cursor: pointer;
+    > input[type="checkbox"] {
+      visibility: hidden;
+    }
+
+    &__inner {
```
sergei-maertens, line 42: `.block__element__subelement` is not proper BEM, any chance you can restructure this?

See also https://nl-design-system.gitlab.io/nl-design-system/componenten/checkbox/index.html please - so far we've been trying to stay a bit consistent with NLDS

src/zac/sass/components/forms/_textarea.scss
```diff
@@ -0,0 +1,26 @@
+.textarea {
```
sergei-maertens, line 1: I know there is a `.input__control--textarea`, defined in `_input.scss` taken from the NLDS - any chance we can re-use that to keep the styles consistent?

sergei-maertens (changes requested):
Great work!

* Checked http://localhost:8000/core/zaken/002220647/ZAAK-2020-0000004015/ as example - and a clickable vs. non-clickable betrokkene row have different paddings which looks a bit weird. Can we fix that? Additionally, looks like having the name be 50% of the width is too much, and the 'Type' column too little, so there's some tweaking in order there too (1920x1080 screen resolution).

* Checkboxes look nice!

* In the modal to request extra details, I'm missing the form label indication that fields are required. Right now you see that the button is disabled, but it's not clear _why_. The existing inputs/labels/wrappers provide that functionality already.

* Noticed the backend API call seems to take a long time (3.53s), but after second call it was clear this was because the API spec wasn't cached yet (740ms)

* Accessibility - the form modal closes on ESC, the results modal does not.

kelvincy:
@sergei-maertens the requested have been fixed.

src/zac/js/components/betrokkenen/BetrokkenenForm.js
```diff
@@ -0,0 +1,239 @@
...
+    params.set('doelbinding', formData.doelbinding);
+    params.set('fields', formData.betrokkeneFields);
+
+    const url = `${endpoint}?${params.toString()}`;
+
+    // send the API call
+    const response = await apiCall(
+        url,
+        {
+            method: 'GET',
```
sergei-maertens, line 115: Discussion on Slack about whether this should be GET or POST

src/zac/sass/components/forms/_textarea.scss
```diff
@@ -0,0 +1,8 @@
+@import "../../lib/bem";
+.input {
+  @include element("control") {
+    @include modifier("text") {
+      font-size: 1rem;
+    }
+  }
+}
```
sergei-maertens, line 8: this conflicts with/overwrites `src/zac/sass/components/forms/_inputs.scss`:

```scss
.input {
  border: 0;
  margin-bottom: 1em;
  padding: 0;
  position: relative;

  @include element('control') {
    border: 0;
    box-shadow: none;
    display: inline-block;
    font-size: 1.1rem;
    margin-right: 2rem;
    min-width: 0;
    width: 100%;

    @include modifier('text') {
      background: #fff;
      border: solid 1px #eee;
      border-radius: 0;
      color: $color-text;
      font: inherit;
      font-size: 1.4rem;
      overflow: visible;
      padding: 0.5rem 1rem;

      &:focus {
        box-shadow: none;
        outline: 2px solid $color-secondary;
      }
    }
  }
}
```

additionally, the `--text` modifier is used for `<input type="text" />` while `--textarea` is used for textareas.

In short - the name of this sass module and the classnames confuse me a lot.

sergei-maertens (changes requested):
The sass components for forms worry me a bit.

I don't think that the current forms are ideal, but it looks like we're now adding conflicting alternative implementations/stylings that are impacting other forms (such as the login form).

Unfortunately we don't have a component library (yet) :(

src/zac/sass/components/forms/_textarea.scss
kelvincy: Fixed

src/zac/js/components/betrokkenen/BetrokkenenForm.js
kelvincy: Changed to POST!

sergei-maertens (approved):

PR #81, opened 2020-09-23
temp auth improvements

### LINKED ISSUE #80: Iterate on temporary access implementation
There are a number of smaller future problems with the current implementation of temporary access to zaken.

- [x] Add the "temporary" aspect. Currently, access is granted and then never taken away. When granting the access, the granter should be able to specify a datetime to expire the access. The permission checks must validate this expiry.
- [x] The handlers machinery is tricky - it works _if_ all the handlers are zaakbehandelaar at the moment of requesting access. However, if someone requests access to a zaak and after that, another user becomes zaakbehandelaar, this second person cannot grant access to the first person because they are not in the `handlers` m2m. The behaviour of who can grant access should probably not rely on a static DB value and do a run-time zaakbehandelaar check instead. This does pose a challenge for the dashboard though.
- [x] Currently, a user can always request access to every zaak they don't have access to. There should be a permission that specifies if a user can request access to the zaak of the particular zaaktype.

annashamray:
Fixes #80


src/zac/accounts/models.py
```diff
@@ -183,5 +187,13 @@ class AccessRequest(models.Model):
...
+    start_date = models.DateField(_("start date"), blank=True, null=True)
+    end_date = models.DateField(_("end date"), blank=True, null=True)
+
+    def clean(self):
+        super().clean()
 
-    objects = AccessRequestQuerySet.as_manager()
+        if self.result and not self.handler:
+            raise ValidationError(
+                _("The result can't be specified without it's handler")
```
sergei-maertens, line 198:
```suggestion
                _("The result can't be specified without its handler")
```

src/zac/core/forms.py
```diff
@@ -362,13 +353,12 @@ def clean(self):
...
 
-        behandelaars = self.get_behandelaars()
-        if behandelaars:
-            request_access.handlers.add(*behandelaars)
+        return super().save()
+
 
-        return request_access
+class DateInputWidget(forms.DateInput):
+    input_type = "date"
```
sergei-maertens, line 361:
```suggestion
```
is unused now

src/zac/core/forms.py
```diff
@@ -380,13 +370,28 @@ class AccessRequestHandleForm(forms.ModelForm):
...
+        submit = self.data.get("submit")
+
+        #  save end date only if the result == approve
+        if not checked:
+            return None
+
+        if submit == AccessRequestResult.approve and not end_date:
+            raise forms.ValidationError(_("End date of the access must be specified"))
+
+        return end_date
```
sergei-maertens, line 393: should be done in `def clean` because you're validating fields that depend on each other

sergei-maertens (approved):
Couple small remarks, but looks great overall!

PR #82, opened 2020-09-23
Set up OO-based authorizations

sergei-maertens:
- [x] Database model for OO
- [x] Assign user to OO
- [x] Limit access to user OO

To the reviewers - I'm sorry for such a massive pull request. It's a complex problem that I'll be making a case about to the rest of the team.

Essentially, with permission checks there are always two branches involved:

- checking the object permissions for a single entity (zaak-detail in our case)
- filtering the data you get back from the API so that only accessible data is displayed

Since the data lives in a remote API, we need to leverage the API endpoint filter parameters to get as close as possible. Therefore, we filter:
- on zaaktype, based on the zaaktypes that the user has access to
- on `rol__betrokkeneIdentificatie__organisatorischeEenheid__identificatie` if relevant, to restrict to zaken that belong to a certain OO

This leads to very complex code where a number of requests are done in parallel based on the configured auth, which populate the zaak-list. The users/permissions leak into that abstraction massively, and it's up for a big refactor because this has reached to point of being easily implemented/used.

The single-entity permission checks are quite straightforward based on `rules`.


src/zac/core/services.py
```diff
@@ -245,23 +245,24 @@ def get_informatieobjecttype(url: str) -> InformatieObjectType:
 ###################################################
 
 
-@cache_result(
-    "zaken:{client.base_url}:{zaaktype}:{max_va}:{identificatie}:{bronorganisatie}",
-    timeout=AN_HOUR,
-)
+# @cache_result(
```
sergei-maertens, line 248: cache is disabled because... it's just not possible anymore to embed all the querystring params in there _and_ be able to figure out all the possible keys to invalidate

src/zac/accounts/models.py
```diff
@@ -87,14 +87,31 @@ class AuthorizationProfile(models.Model):
     """
 
     uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
-    name = models.CharField(_("naam"), max_length=255)
+    name = models.CharField(
+        _("naam"),
```
damm89, line 91: is het met de vertaaltabel niet handiger/vollediger om Engelse woorden te gebruiken? _("naam") -> _("name")

src/zac/accounts/models.py
```diff
@@ -87,14 +87,31 @@ class AuthorizationProfile(models.Model):
...
         help_text=_(
             "Selecting multiple sets makes them add/merge all the permissions together."
         ),
     )
+    oo = models.ForeignKey(
+        "organisatieonderdelen.OrganisatieOnderdeel",
+        null=True,
+        blank=True,
+        on_delete=models.PROTECT,
+        verbose_name=_("organisatieonderdeel"),
```
damm89, line 109: ""

src/zac/contrib/organisatieonderdelen/models.py
```diff
@@ -0,0 +1,15 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+
+class OrganisatieOnderdeel(models.Model):
+
+    name = models.CharField(_("name"), max_length=100)
+    slug = models.SlugField(_("slug"), max_length=24, unique=True)
+
```
damm89, line 9: ""

src/zac/core/services.py
damm89: what do you think the performance impact of this is?

src/zac/core/rules.py
```diff
@@ -49,27 +32,89 @@ def __getattr__(self, attr: str):
...
 
 
+def _get_oos_from_zt_perms(user_perms: UserPermissions, zaaktype: str) -> set:
+    if user_perms.user.is_superuser:
+        return {None}
+    perm_oos = [
+        zt_perm.oos
+        for zt_perm in user_perms.zaaktype_permissions
+        if zt_perm.contains(zaaktype) and zt_perm.permission == zaken_inzien.name
+    ]
```
annashamray, line 73: Should filtering also take into account `zaak.va` ? For example, let's imagine two auth profiles for user Bob:
* profile A with `oo` = "oo1" and 1 permission set for "zaaktype1"  with `va` = "openbare"
* profile B with `oo` = "oo2" and 1 permission set for the same "zaaktype1" with `va`= "zeer_geheim"

And now we are checking permissions for the zaak with `zaaktype` = zaaktype1 and `va` = "zeer_geheim"

In current implementation both "oo1" and "oo2" would be selected here as relevant oos, but is it correct?

src/zac/core/tests/test_zaak_detail.py
```diff
@@ -803,3 +807,183 @@ def test_claim_task_with_permission(self):
...
+        )
+        auth_profile = perm_set.authorizationprofile_set.get()
+        auth_profile.oo = oo
+        auth_profile.save()
+
+        # mock out all the other calls - we're testing auth here
+        with mock_zaak_detail_context():
+            response = self.app.get(self.url, user=self.user, status=403)
+
+        self.assertEqual(response.status_code, 403)
```
annashamray, line 989: Could you also add tests for zaak detail page?
* user is a superuser
* no oo is set

src/zac/core/services.py
sergei-maertens: It slows down the page when there are no new results - zaak-creation for example already resulted in the cache being invalidated to get the new results.

src/zac/accounts/models.py
sergei-maertens: I have no idea what the "correct" English translation would be for this, so I'll stick with the Dutch term here.

src/zac/accounts/models.py
sergei-maertens: yep, changing it

src/zac/contrib/organisatieonderdelen/models.py
sergei-maertens: `slug` is the English term, not sure what else can be changed here

src/zac/core/rules.py
sergei-maertens: In this case not, to not complicate it too much. This function is called in two places:

1. `get_zaken` to build the list of zaken, which includes the `maxVertrouwelijkheidsaanduiding` query parameter on the API level
2. `test_oo_allowlist`, which is invoked after the object level permissions based on VA/zaaktype/permission have passed.

It _is_ dangerous to apply this as first function without additional checks. One of the reasons I want to take a proper look at the auth protections in place now, this needs refactoring because it has complexity that was not anticipated.

src/zac/core/tests/test_zaak_detail.py
sergei-maertens: done

src/zac/core/rules.py
sergei-maertens: I managed to resolve it without getting stuff too complex :-) build is passing locally now, so I'll wait for Travis and then merge

Issue #83, opened 2020-09-24
Implement authorization on document types and security levels.


sergei-maertens:
Applies to: 
- zaak-detail page
- download of document view

On the zaak-detail page, the related documents are displayed. These are retrieved via the `zaakinformatieobject_list` API operation. Each document ("enkelvoudiginformatieobject") is of a certain document type ("informatieobjecttype") and has a certain security level ("vertrouwelijkheidaanduiding"). This is entirely parallel to `zaak` with `zaak.zaaktype` and `zaak.vertrouwelijkheidaanduiding`.

The ZAC currently has a permission system where an Authorization Profile (`accounts` app) is made up of one or multiple permission sets. Users can have one or more authorization profiles, which grant them the rights from the "embedded" permission sets.

A permission set is currently defined as a collection of permissions on zaken of set of selected zaaktypen where the confidentiality of the zaak is below a certain maximum confidentiality. Currently, if you are allowed to download documents in zaken of a given zaaktype, you can do so for all document types.

This functionality needs to be extended to also (optionally) limit the informatieobjecttypes and max confidentiality level. This means that within the zaak-detail page, only the documents of the zaak are displayed where:
- the document type is on the allow-list
- the `document.vertrouwelijkheidaanduiding` is not more confidential than the maximum configured one.

Note that if the catalogus is selected, and no (sub)set of informatieobjecttypen is selected, this means that _all_ informatieobjecttypen are selected. This is the same as with zaaktypen.

Acceptance criteria:

- [x] When creating or editing a permission set, I can select the catalogus to use for informatieobjecttypen
- [x] When creating or editing a permission set, after selecting the catalogus, I can select a set of informatieobjecttypen to grant access to
- [x] When creating or editing a permission set, I can select a `max_vertrouwelijkheidaanduiding` for documents from a list of options (there is an enum)
- [x] When a user views a zaak
    - [x] The documents related to the zaak are filtered on informatieobjecttype
    - [x] The documents related to the zaak are filtered on vertrouwelijkheidaanduiding
- [x] When a user guesses or knows the download URL of a document, the permissions are checked based on:
    - [x] informatieobjecttype
    - [x] vertrouwelijkheidaanduiding

PR #84, opened 2020-09-29
Issue 83 document authorisation

SilviaAmAm:
Fixes issue #83


sergei-maertens (commented):
I'm reluctantly approving this.

The `user` abstraction leaking into `zac.core.services` bothers me - this module is supposed to just implement the interface to the API and not concern itself with request/response cycle. The functions are supposed to be used outside of request-response cycles too, where no users exists. I understand why it was done, and the whole permission handling machinery needs to be refactored with the lessons learned.

But in general, try to keep a clear separation of concerns.

There are a couple minor suggested changes I want to see addressed before merging this though.

PR #85, opened 2020-09-29
Added feature that allows for end user to provide a review request wi…

damm89:
…th an explanatory note

GemeenteUtrecht/ZGW#547

### DIFF
--- a/src/zac/contrib/kownsl/data.py
+++ b/src/zac/contrib/kownsl/data.py
@@ -26,7 +26,7 @@ class ReviewRequest(Model):
     num_advices: int
     num_approvals: int
     num_assigned_users: int
-    toelichting: str
+    toelichting: str = ""
 
     def get_review_type_display(self):
         return KownslTypes.labels[self.review_type]
@@ -74,4 +74,4 @@ class Approval(Model):
     created: datetime
     author: Author
     approved: bool
-    toelichting: str
+    toelichting: str = ""
--- a/src/zac/contrib/kownsl/data.py
+++ b/src/zac/contrib/kownsl/data.py
@@ -26,6 +26,7 @@ class ReviewRequest(Model):
     num_advices: int
     num_approvals: int
     num_assigned_users: int
+    toelichting: str
 
     def get_review_type_display(self):
         return KownslTypes.labels[self.review_type]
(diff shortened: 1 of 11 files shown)


src/zac/contrib/kownsl/data.py
```diff
@@ -26,6 +26,7 @@ class ReviewRequest(Model):
     num_advices: int
     num_approvals: int
     num_assigned_users: int
+    toelichting: str
```
annashamray, line 29: if ZAAK has already some review requests without `toelichting` this will crush the page. Could you add a default value, please?

src/zac/contrib/kownsl/data.py
```diff
@@ -73,3 +74,4 @@ class Approval(Model):
     created: datetime
     author: Author
     approved: bool
+    toelichting: str
```
annashamray, line 77: Could you add a default value, please?

annashamray (changes requested):
Looks good (especially tests) but I have some concerns.

In the Michelle's issue she asks to add `toelichting` only for `Approval`
In this PR `toelichting` is added to both models - `ReviewRequest` and `Approval`, which was not specified in the requirements. But perhaps I misunderstood the issue. We can have a call to clear this out

src/zac/contrib/kownsl/data.py
damm89: You're right!

src/zac/contrib/kownsl/data.py
damm89: Yes :)

annashamray (approved):
I don't know why we haven't used grid before in this project, maybe @sergei-maertens could know it. Perhaps we can discuss and create some style guidelines for this project in the nearest future to make all forms consistent

sergei-maertens (approved):

sergei-maertens:
> I don't know why we haven't used grid before in this project

It's not supported by Internet Explorer, see https://caniuse.com/css-grid

I think IE is no longer used for demo's, so we should be fine now.

sergei-maertens:
Double checked, but because of the default values this can be merged before the Kownsl stuff is merged (it just means that if we deploy it now that the information is "lost")

PR #87, opened 2020-10-05
Added new variable to be passed to bpmn process and changed variable …

damm89:
…names

GemeenteUtrecht/ZGW#494


sergei-maertens (approved):
Looks good, will just need to update the process models before deploying!

sergei-maertens:
I see https://github.com/GemeenteUtrecht/ZGW/pull/563 now - will review!

PR #89, opened 2020-10-07
Feature/django hijack

### LINKED ISSUE #88: Install and enable django-hijack
To be used in the BBV demo's, so high prio.

- [ ] As an admin user I can hijack other users
    - [x] As a non-superuser admin user I can hijack other users
- [x] While hijacking, I can release the user account again from the main pages

annashamray:
Fixes #88

### DIFF
--- a/src/zac/accounts/admin.py
+++ b/src/zac/accounts/admin.py
@@ -19,7 +19,7 @@ class UserAuthorizationProfileInline(admin.TabularInline):
 
 
 @admin.register(User)
-class _UserAdmin(UserAdmin, HijackUserAdminMixin):
+class _UserAdmin(HijackUserAdminMixin, UserAdmin):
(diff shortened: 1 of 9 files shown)


src/zac/accounts/admin.py
```diff
@@ -17,7 +19,8 @@ class UserAuthorizationProfileInline(admin.TabularInline):
 
 
 @admin.register(User)
-class _UserAdmin(UserAdmin):
+class _UserAdmin(UserAdmin, HijackUserAdminMixin):
```
sergei-maertens, line 22:
```suggestion
class _UserAdmin(HijackUserAdminMixin, UserAdmin):
```
mixins should always go before the base class(es) because of [MRO](https://www.python.org/download/releases/2.3/mro/)

sergei-maertens (approved):
Minor remark, if you fix that you can merge this PR!

PR #92, opened 2020-10-09
Feature/elastic search

annashamray:
Implemented changes:
* `index_zaken` management command is added to the project, which indexes all zaken with nested rollen
*  ES documents are being updated when notifications from zaken are received
* `search_zaken` function is created but not used

TODO:
* after merging OO-based auth rol filters can be included into `search_zaken`function
* after implementing batch endpoint in Openzaak `search_zaken` can be used to list all relevant zaken for a user

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,12 @@ env:
   - DB_USER=postgres
   - IS_HTTPS=no
 
+before_install:
+  - curl https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.9.2-amd64.deb -o elasticsearch.deb
+  - sudo dpkg -i --force-confnew elasticsearch.deb
+  - sudo chown -R elasticsearch:elasticsearch /etc/default/elasticsearch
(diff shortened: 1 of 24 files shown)


.travis.yml
```diff
@@ -18,6 +19,16 @@ env:
...
 
+before_install:
+  - curl https://artifacts.elastic.co/downloads/elasticsearch/elasticsearch-7.9.2-amd64.deb -o elasticsearch.deb
+  - sudo dpkg -i --force-confnew elasticsearch.deb
+  - sudo chown -R elasticsearch:elasticsearch /etc/default/elasticsearch
+  - sudo service elasticsearch restart
+
+# ElasticSearch takes few seconds to start
+before_script:
+  - sleep 10
```
sergei-maertens, line 30: we could maybe include wait-for-it.sh and let that check for the ES host and port?

sergei-maertens (approved):
Looks good, excited to start trying this out, but as you said, first need to sort that OO-branch out.

PR #96, opened 2020-10-16
Specifying permissions on each informatieobjecttype

SilviaAmAm:


sergei-maertens:
To recap what was said on Slack - I tested quickly (and did run some chrome debugger inspections) how it looks in the user interface and this is indeed the direction it should go.

Styling wise there's a lot to gain, I'll take that on me probably or assign it to Kelvin. I haven't done a code-review yet to check complexity, which I'll do tomorrow.

src/zac/js/components/informatieobjecttype-forms/informatieobjecttype-form.js
```diff
@@ -0,0 +1,201 @@
...
+        if (event.target.value !== '') {
+            this.fetchInformatieobjecttypen(event.target.value);
+        }
+    }
+
+    fetchInformatieobjecttypen (catalogUrl) {
+        const apiURL = "/accounts/permission-sets/informatieobjecttypes?catalogus=" + catalogUrl;
+        window.fetch(apiURL).then(
+            response => response.json()
+        ).then(
```
sergei-maertens, line 117: You can also use `async/await` syntax to make this look like regular synchronous code, while still using promises under the hood:

```jsx
import {get} from '../../utils/fetch';

const IOTEndpoint = '/accounts/permission-sets/informatieobjecttypes';

const informatieobjectTypes = await get(IOTEndpoint, {catalogus: catalogUrl});
```

src/zac/js/components/informatieobjecttype-forms/informatieobjecttype-form.js
```diff
@@ -0,0 +1,201 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {FormSet} from './formset';
+import {Select, FormSetSelect} from '../forms/Select';
+import {CheckboxInput, HiddenInput} from "../forms/Inputs";
+
+
+
+class InformatieobjecttypePermissionForm extends React.Component {
```
sergei-maertens, line 9: In general, I see a lot of component state management here that is *probably* not needed.

React has two types of inputs: controlled v.s. uncontrolled. The former is what you mostly see in SPA, the latter is particularly suited to just getting the DOM in the shape necessary to submit to the backend. These essentially operate using the `value` or `defaultValue` prop on form components (see also https://scriptverse.academy/tutorials/reactjs-defaultvalue-value.html).

Long story short - if we make sure that everything is output into `defaultValue` by using the right props, the browser will actually do the heavy lifting for us. Checking a checkbox is exactly that, the browser knows the state of that. Changing a dropdown -> also essentially managed by the browser by default.

So I think we can get rid of most of the state management in these components and make them stateless (or whatever state is there, is there in the browser itself).

sergei-maertens (changes requested):

src/zac/js/components/informatieobjecttype-forms/formset.js
```diff
@@ -0,0 +1,66 @@
...
+        minNum: PropTypes.number.isRequired,
+        maxNum: PropTypes.number.isRequired,
+    }).isRequired,
+    renderForm: PropTypes.func.isRequired, // a render prop
+    formData: PropTypes.arrayOf(PropTypes.object).isRequired,
+    emptyFormData: PropTypes.arrayOf(PropTypes.object),
+};
+
+
+export { FormSet };
```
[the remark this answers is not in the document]
SilviaAmAm: I wanted to control the data that goes into the extra forms: in the formset in `src/zac/js/components/formsets/FormSet.js` the extra forms are empty.

I made a new one mostly because I wasn't sure this was the right way to go and I didn't want to break other things in the process of trying it out :sweat_smile: But if it was okay then I'll to update the existing formset implementation.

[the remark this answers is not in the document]
sergei-maertens: > I wanted to control the data that goes into the extra forms: in the formset in src/zac/js/components/formsets/FormSet.js the extra forms are empty.

hrm, I *think* you can actually control that by passing it as an extra Object in the regular `formData` array. For the backend, it doesn't matter if it's an `extra` form or not - it's all treated the same. What matters is the hidden `id` field, which determines if it's an update or a create.

src/zac/accounts/api/views.py
```diff
@@ -0,0 +1,37 @@
...
+    authentication_classes = [SessionAuthentication]
+    permission_classes = [IsAuthenticated]
+
+    def get(self, request):
+        """Return the informatieobjecttypen for a catalogus"""
+        catalogus_url_serializer = CatalogusURLSerializer(
+            data={"url": request.GET.get("catalogus")}
+        )
+        if not catalogus_url_serializer.is_valid():
+            raise ValidationError
```
sergei-maertens, line 22:
```suggestion
        catalogus_url_serializer.is_valid(raise_exception=True)
```

src/zac/accounts/views.py
```diff
@@ -42,23 +54,101 @@ class PermissionSetsView(LoginRequiredMixin, ListView):
...
 
 class PermissionSetDetailView(LoginRequiredMixin, DetailView):
     queryset = PermissionSet.objects.prefetch_related("authorizationprofile_set")
     context_object_name = "permission_set"
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        informatieobjecttype_permissions = (
+            InformatieobjecttypePermission.objects.filter(
+                permission_set=context["object"]
```
sergei-maertens, line 99:
```suggestion
                permission_set=self.object
```
`self.object` is set at the beginning of the `def get` in `DetailView`. It's the result of `self.get_object()`. Pulling stuff from the context while it's available as a view attribute seems a bit backwards.

src/zac/core/rules.py
```diff
@@ -2,7 +2,10 @@
 from datetime import date
 from typing import Optional, Set, Union
 
+from django.core.exceptions import PermissionDenied
+
 import rules
+from odf.office import Document
```
sergei-maertens, line 8: wrong auto-import?

src/zac/core/views/utils.py
```diff
@@ -12,3 +20,69 @@ def get_zaak_from_query(request: HttpRequest, param: str = "zaak") -> Zaak:
...
+    )
+
+    if not user_iot_permissions.exists():
+        raise PermissionDenied(error_msg)
+
+    # The user has permissions to the catalog in question.
+    # Now need to check whether they have the right confidentiality level
+
+    document_va = VertrouwelijkheidsAanduidingen.get_choice(
+        document.informatieobjecttype.vertrouwelijkheidaanduiding
```
sergei-maertens, line 51:
```suggestion
        document.vertrouwelijkheidaanduiding
```
The document itself has a VA - which is by default the value from the IOT, but may be overridden.

src/zac/core/views/utils.py
```diff
@@ -12,3 +20,69 @@ def get_zaak_from_query(request: HttpRequest, param: str = "zaak") -> Zaak:
 
     zaak = get_zaak(zaak_url=zaak_url)
     return zaak
+
+
+def check_document_permissions(document: Document, user: User):
```
sergei-maertens, line 25: I'm expecting a performance hit from the amount of queries happening in there, especially if it's being looped in `filter_documenten_for_permissions`. However, let's first verify the correctness and then tackle performance.

sergei-maertens (changes requested):
I'll to through the frontend myself for a bit, especially with regard to:

1. Using a single `FormSet` class and organizing things so that it's possible to do so
2. Seeing how much state I can remove and make the forms uncontrolled components.

sergei-maertens:
Ran the CI steps locally on machine because the Travis backlog is horrendous again... Everything passes now.

sergei-maertens (approved):

PR #98, opened 2020-10-21
Feature/elastic search on oo

annashamray:
Changes:
* include filtering on `oo` to search function


sergei-maertens (approved):
More basic plumbing to start using this - the next step is hooking this up to the views, for which I've started an experimental branch. Tested the indexing and search functionality, and this seems very promising w/r to performance!

PR #99, opened 2020-10-21
Feature/termijnbewaking review request

kelvincy:


src/zac/accounts/api/viewsets.py
```diff
@@ -7,9 +7,23 @@
...
     search_fields = ["username", "first_name", "last_name"]
+
+    def get_queryset(self):
+        filter_users = self.request.query_params.get("filter_users", "")
+        queryset = User.objects.filter(is_active=True)
+        if filter_users:
+            # Custom cleaning of query_parameter
+            # TODO find better implementation
+            filter_users = filter_users.split(",")
+            include = self.request.query_params.get("include", False)
```
sergei-maertens, line 23: Should be a django-filter implementation, pretty sure it has built in filter types for this.

src/zac/core/forms.py
```diff
@@ -323,37 +323,95 @@ class SelectUsersReviewRequestForm(forms.Form):
...
 
 class BaseReviewRequestFormSet(BaseTaskFormSet):
-    def is_valid(self):
-        super().is_valid()
+    def deadlines_validation(self) -> bool:
+        """
+        Validate that deadlines per step are monotonic increasing
+        """
+        valid = True
+        deadline_old = date(1, 1, 1)
```
sergei-maertens, line 340:
```suggestion
        deadline_old = date.today()
```
Deadlines should be in the future imo, so we can use a realistic base-line with today's date?

src/zac/core/forms.py
```diff
@@ -323,37 +323,95 @@ class SelectUsersReviewRequestForm(forms.Form):
...
-            if not form.has_changed():
-                form.add_error(None, _("Please select at least 1 advisor."))
-                return False
+            deadline_new = form.cleaned_data["deadline"]
+            if deadline_new and not deadline_new > deadline_old:
+                form.add_error(
+                    None,
+                    _(
+                        f"Please select a date greater than {deadline_old.strftime('%Y-%m-%d')}"
+                    ),
```
sergei-maertens, line 348: This does not work with translatable strings, as the 'template' contains a dynamic value. Please use placeholders - in general, f-strings are not suitable for translatable content.

```suggestion
                    _("Please select a date greater than {minimum_date}").format(minimum_date=deadline_old.strftime('%Y-%m-%d'))
```

src/zac/core/forms.py
```diff
@@ -323,37 +323,95 @@ class SelectUsersReviewRequestForm(forms.Form):
...
+        valid = True
+        deadline_old = date(1, 1, 1)
         for form in self.forms:
-            if not form.has_changed():
-                form.add_error(None, _("Please select at least 1 advisor."))
-                return False
+            deadline_new = form.cleaned_data["deadline"]
+            if deadline_new and not deadline_new > deadline_old:
+                form.add_error(
+                    None,
```
sergei-maertens, line 345:
```suggestion
                    "deadline",
```
You can attach the error directly to the form.deadline field instead of putting it in the `form.non_field_errors`

src/zac/core/forms.py
```diff
@@ -323,37 +323,95 @@ class SelectUsersReviewRequestForm(forms.Form):
...
+    def unique_user_validation(self) -> bool:
+        """
+        Validate that users are unique and that at least 1 user is selected per step
+        """
+        valid = True
+        users_list = []
+        for form in self.forms:
+            users = form.cleaned_data["kownsl_users"]
+            if not users:
+                form_add_error(None, _("Please select at least 1 advisor."))
```
sergei-maertens, line 362: advisor/approver - seems to be context dependent, so I'd make this dynamic to provide the correct feedback. You can also attach the error to the `users` field of the form rather than putting it in the non-field errors.

src/zac/core/forms.py
```diff
@@ -323,37 +323,95 @@ class SelectUsersReviewRequestForm(forms.Form):
...
+                valid = False
+
+            if any([user in users_list for user in users]):
+                form.add_error(None, _("Please select unique advisors."))
```
sergei-maertens, line 366: Similar remarks to the above.

src/zac/core/forms.py
```diff
@@ -323,37 +323,95 @@ class SelectUsersReviewRequestForm(forms.Form):
...
+    def is_valid(self) -> bool:
+        valid = super().is_valid()
+        # make sure deadlines are monotonic increasing and all users are unique
+        return all(
+            [
+                valid,
+                self.deadlines_validation(),
+                self.unique_user_validation(),
+            ]
+        )
```
sergei-maertens, line 381: Should be implemented in the `FormSet.clean` method, see https://docs.djangoproject.com/en/2.2/topics/forms/formsets/#custom-formset-validation

sergei-maertens (changes requested):

src/zac/accounts/api/tests/test_viewsets.py
```diff
@@ -0,0 +1,47 @@
...
+        self.client.force_authenticate(user=self.superuser)
+        self.url = reverse("accounts:users-list")
+
+    def test_view_url_exists(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
+
+    def test_view_search_users(self):
+        search_url = self.url + "?search=h"
+        response = self.client.get(search_url)
```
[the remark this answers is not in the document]
damm89: Got it

src/zac/accounts/api/viewsets.py
damm89: Alright, will look at it and change it in the BPTL as well.

src/zac/core/forms.py
damm89: Alright

src/zac/core/forms.py
damm89: well, the current frontend doesn't allow for the extra forms to persist once submitted. Attaching it to the field of a non-existing form doesn't seem to have any real benefits?

damm89 (approved):

sergei-maertens (approved):

src/zac/core/forms.py
sergei-maertens: Hrm, current solution seems sufficient then. I think UI-wise there's something to gain, but it would be better suited to include that in the form-data that's passed to the frontend and I don't think we should delay this feature any longer because of that, it can easily be tackled later.

sergei-maertens:
Ok, verified the CI steps locally on my machine as the Travis build queue is massive again:

- [x] black ok
- [x] isort ok
- [x] docker build ok
- [x] tests pass

So, I'm merging this into develop.

PR #101, opened 2020-10-26
Initial experiment using ES index for permission-related list queries

sergei-maertens:
Related issue: https://github.com/GemeenteUtrecht/ZGW/issues/551

Changes:

* "Alle zaken" are queried using ES
* `get_zaken` function is refactored: query params from the filter form and filters based on permissions are separated. 

Todo (in the next PRs since we try to keep PRs small):
* use ES for querying zaken in the dashboard
* validate filter params before usin them in `get_zaken`, because it couldn't use some arbitrary parameters anymore

### DIFF
--- a/src/zac/core/services.py
+++ b/src/zac/core/services.py
@@ -31,6 +31,7 @@ from zgw_consumers.service import get_paginated_results
 from zac.accounts.datastructures import VA_ORDER
 from zac.accounts.permissions import UserPermissions
 from zac.contrib.brp.models import BRPConfig
+from zac.elasticsearch.searches import search
 from zac.utils.decorators import cache as cache_result
 from zgw.models import InformatieObjectType, StatusType, Zaak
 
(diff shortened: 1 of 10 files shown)


src/zac/core/services.py
```diff
@@ -392,6 +392,74 @@ def get_zaken(
...
+    for zaak in zaken:
+        zaak.zaaktype = _base_zaaktypen[zaak.zaaktype]
+
+    # sort results by startdatum / registratiedatum / identificatie
+
+    zaken = sorted(
+        zaken,
+        key=lambda zaak: (zaak.registratiedatum, zaak.startdatum, zaak.identificatie),
+        reverse=True,
+    )
```
sergei-maertens, line 458: This sort should probably happen in ES and not on the results of ES. We rely on the order of ES results here - `zaak_urls` is already something that came from ES.

src/zac/elasticsearch/tests/utils.py
```diff
@@ -15,7 +20,17 @@ def refresh_index(self):
...
+    def add_rol_to_document(self, rol):
+        if not isinstance(rol, Rol):
+            rol = factory(Rol, rol)
+        append_rol_to_document(rol)
+
     def setUp(self):
         super().setUp()
 
-        self._clear_index()
+        self.clear_index()
```
sergei-maertens, line 36: slightly more robust:

```suggestion
        self.addCleanup(self.clear_index)
```

sergei-maertens:
Tests verified locally on my machine, this looks great!

PR #103, opened 2020-10-29
Feature/serieel accorderen improvements

kelvincy:
Feature: aansluiting op de gewijzigde BE van serieel accorderen om al geselecteerde gebruikers te excluden van de search results.

### DIFF
--- a/src/zac/js/components/formsets/FormSet.js
+++ b/src/zac/js/components/formsets/FormSet.js
@@ -47,7 +47,7 @@ const FormSet = ({
     const extraForms = Array(extra).fill().map(
         (_, index) => (
             <PrefixContext.Provider key={existingCount + index} value={getPrefix(existingCount + index)}>
-                <RenderForm index={existingCount + index} totalStepsIndex={extra} data={{}} onDelete={onDelete} />
+                <RenderForm title={configuration.title} index={existingCount + index} totalStepsIndex={extra} data={{}} onDelete={onDelete} />
             </PrefixContext.Provider>
         ),
     );
--- a/src/zac/js/components/formsets/FormSet.js
+++ b/src/zac/js/components/formsets/FormSet.js
@@ -40,14 +40,14 @@ const FormSet = ({
     const forms = formData.map(
         (data, index) => (
             <PrefixContext.Provider key={index} value={getPrefix(index)}>
-                <RenderForm title={configuration.title} index={index} totalStepsIndex={extra} data={data} onDelete={onDelete} />
+                <RenderForm index={index} totalStepsIndex={extra} data={data} onDelete={onDelete} />
             </PrefixContext.Provider>
         ),
     );
     const extraForms = Array(extra).fill().map(
         (_, index) => (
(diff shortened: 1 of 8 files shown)


src/zac/js/utils/fetch.js
```diff
@@ -1,34 +1,38 @@
...
-const get = async (url, params={}) => {
+const get = async (url, params = {}, multiParams = []) => {
+    let searchParams;
     if (Object.keys(params).length) {
-        const searchparams = new URLSearchParams(params);
-        url += `?${searchparams}`;
+        searchParams = new URLSearchParams(params);
+    }
+    if (multiParams.length > 0) {
+        multiParams.map((param) => searchParams.append(
```
sergei-maertens, line 18: This will break if `params` is empty, but `multiParams` isn't because `searchParams` is `undefined`.

src/zac/js/utils/fetch.js
```diff
@@ -1,34 +1,38 @@
...
+            Object.keys(param)[0], param[Object.keys(param)[0]],
+        ));
```
sergei-maertens, line 20: How about a bit more readable version? I had to really double-check with https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/append to figure out what exactly is happening in the code :) 

```suggestion
        multiParams.map((param) => {
            const paramName = Object.keys(param)[0]; // param={foo: bar}
            searchParams.append(paramName, param[paramName]);
        });
```

sergei-maertens (changes requested):
Not sure about adding the `title` configuration key, as I'd like that FormSet configuration to mirror the Django backend formsets. I *think* it'd be better to use a React context to provide this without having to pass it along as props to the individual forms.

src/zac/js/components/formsets/FormSet.js
```diff
@@ -82,6 +82,7 @@ FormSet.propTypes = {
         extra: PropTypes.number.isRequired,
         minNum: PropTypes.number.isRequired,
         maxNum: PropTypes.number.isRequired,
+        title: PropTypes.string,
```
sergei-maertens, line 85: I think this proptype should be removed alltogether?

src/zac/js/utils/fetch.js
```diff
@@ -1,34 +1,40 @@
...
-
-const get = async (url, params={}) => {
+const get = async (url, params = {}, multiParams = []) => {
+    const searchParams = new URLSearchParams();
     if (Object.keys(params).length) {
-        const searchparams = new URLSearchParams(params);
-        url += `?${searchparams}`;
+        const paramName = Object.keys(params)[0];
+        searchParams.append(paramName, params[paramName]);
+    }
```
sergei-maertens, line 17: this does not work if `params` contains multiple keys.

shouldn't this be something like `searchParams.append(params)` (untested)?

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #105, opened 2020-10-30
Feature/es search for dashboard

annashamray:
Changes:
* use ES for querying zaken in the dashboard
* check search parameters for zaken

### DIFF
--- a/src/zac/core/services.py
+++ b/src/zac/core/services.py
@@ -294,110 +294,6 @@ def _find_zaken(
     return _zaken
 
 
-def get_zaken(
-    user_perms: UserPermissions,
-    zaaktypen: List[str] = None,
-    identificatie: str = "",
-    bronorganisatie: str = "",
-    find_all=False,
-    **query_params,
-) -> List[Zaak]:
-    """
-    Fetch all zaken from the ZRCs.
-
-    Only retrieve what the user permissions dictate.
-    """
-    _base_zaaktypen = {zt.url: zt for zt in get_zaaktypen(user_perms)}
-    allowed_zaaktypen = _base_zaaktypen
-
-    if user_perms.user.is_superuser:
-        query_zaaktypen = zaaktypen or [""]
-    else:
-        query_zaaktypen = (
-            allowed_zaaktypen.keys()
-            if not zaaktypen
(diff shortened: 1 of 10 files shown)


src/zac/core/services.py
```diff
@@ -392,6 +392,75 @@ def get_zaken(
...
+
+    # ES search
+    from zac.elasticsearch.searches import search
+
+    zaak_urls = search(size=size, **find_kwargs)
+
+    def _get_zaak(zaak_url):
+        return get_zaak(zaak_url=zaak_url)
+
+    with parallel(max_workers=10) as executor:
```
damm89, line 445: max_workers=10 confuses me a little bit. Why 10? We talked about it briefly during our meeting but I'm not sure something came out of it.

I know in Docker you can't fetch the available CPU count by using os.cpu_count or something similar, but we might be able to make a workaround using openjdk... Idk. Maybe this works as it is for now perfectly fine. :-)

annashamray: Tbh I haven't put too many thoughts into it and just copied the existing use of `parallel(max_workers=10)` in ZAC (for example in `core/services/search_zaak_for_related_object` function)
Maybe, we can just exclude it since there are also a lot of uses of `parallel` in ZAC without specifying `max_workers`
Since Python 3.5 the default amount of workers = CPUs * 5, so in any case, it should be enough I guess? 

@sergei-maertens could you explain the reason behind using `parallel(max_workers=10)` in ZAC, please?

src/zac/elasticsearch/searches.py
```diff
@@ -1,42 +1,80 @@
...
+from elasticsearch_dsl.query import Bool, Nested, Range, Term, Terms
 from zgw_consumers.api_models.constants import VertrouwelijkheidsAanduidingen
 
 from .documents import ZaakDocument
 
 
-def search_zaken(
-    size=25,
-    zaaktypen=None,
+def search(
```
damm89, line 12: very nice

src/zac/core/services.py
sergei-maertens: ehm... just a magic number, really. Didn't give that any thought

sergei-maertens (approved):

PR #111, opened 2020-11-05
Hook into Kownsl notifications

sergei-maertens:
This is the handler for the feature introduced in the latest Kownsl version: https://github.com/GemeenteUtrecht/kownsl/pull/19

### DIFF
--- a/src/zac/contrib/kownsl/views.py
+++ b/src/zac/contrib/kownsl/views.py
@@ -1,0 +1,12 @@
+from zac.notifications.views import NotificationCallbackView
+
+
+class KownslNotificationCallbackView(NotificationCallbackView):
+    def handle_notification(self, data: dict):
+        # just to make sure, shouldn't happen with our URL routing
+        if not data["kanaal"] == "kownsl":
+            return
+
+        import bpdb
+
+        bpdb.set_trace()
--- a/src/zac/contrib/kownsl/views.py
+++ b/src/zac/contrib/kownsl/views.py
@@ -1,4 +1,12 @@
+import logging
+
+from django_camunda.api import complete_task
+from django_camunda.client import get_client as get_camunda_client
+from zgw_consumers.models import Service
+
 from zac.notifications.views import NotificationCallbackView
+
+logger = logging.getLogger(__name__)
 
 
 class KownslNotificationCallbackView(NotificationCallbackView):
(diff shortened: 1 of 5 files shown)


src/zac/contrib/kownsl/views.py
```diff
@@ -0,0 +1,53 @@
...
+from django_camunda.api import complete_task
+from django_camunda.client import get_client as get_camunda_client
+from zgw_consumers.models import Service
+
+from zac.notifications.views import NotificationCallbackView
+
+logger = logging.getLogger(__name__)
+
+
+class KownslNotificationCallbackView(NotificationCallbackView):
```
damm89, line 12: Is using NotificationCallbackView truly warranted here? Instead of using protected functions and then inherit maybe subclass NotificationCallbackView into a Base/GenericCallbackView and relevant other classes?

src/zac/contrib/kownsl/views.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @staticmethod
+    def _handle_review_submitted(data: dict):
+        client = Service.get_client(data["hoofd_object"])
+
+        logger.debug("Retrieving review request %s", data["hoofd_object"])
+        review_request = client.retrieve("reviewrequest", url=data["hoofd_object"])
+
+        # look up and complete the user task in Camunda
+        assignee = data["kenmerken"]["author"]
```
damm89, line 29: I'm guessing you want this to break if no author key is found?

src/zac/notifications/tests/test_kownsl_review_submitted.py
```diff
@@ -0,0 +1,113 @@
...
+            json=[],
+        )
+
+        with patch("zac.contrib.kownsl.views.complete_task") as mock_complete:
+            response = self.client.post(self.endpoint, NOTIFICATION)
+
+        self.assertEqual(response.status_code, status.HTTP_204_NO_CONTENT)
+
+        self.assertEqual(m.last_request.method, "GET")
+        mock_complete.assert_not_called()
```
damm89, line 113: Ideally I'd like to have an exception test in case input data is crap because a change of serializer or w/e.

src/zac/contrib/kownsl/views.py
sergei-maertens: Good point, will refactor that!

src/zac/contrib/kownsl/views.py
sergei-maertens: Exactly :) needs to end up in Sentry where we see the relevant data and can then resolve it manually if needed.

src/zac/notifications/tests/test_kownsl_review_submitted.py
sergei-maertens: That would mean the Notifications API is not behaving according to the standard, which is "not our problem", no?

damm89: Not entirely but it would be nice to see which exceptions would be thrown.

damm89 (approved):

src/zac/notifications/tests/test_kownsl_review_submitted.py
sergei-maertens: I don't think this is a useful addition, there can be a vast range of exceptions but they'll all end up in Sentry which is used for error monitoring. If it's something that can/should be handled on our end, that would be the time to write regression tests. However, in its current form, our callback endpoint would just throw a HTTP 400 towards the notifications API.

damm89: Fair enough

PR #113, opened 2020-11-09
Request access permissions bugfix

sergei-maertens:


src/zac/core/rules.py
```diff
@@ -77,15 +77,14 @@ def _get_oos_from_zt_perms(
         for zt_perm in user_perms.zaaktype_permissions
```
damm89, line 77: Poorly legible.

damm89 (approved):

src/zac/core/rules.py
sergei-maertens: Discussed on Slack - we're not sure how long this code will even still exist given that we don't want to hard-couple with OO's. For the time being we leave this as is.

Issue #109, opened 2020-11-03
Add functionality to relate zaken from the Zaak Detail page


SilviaAmAm:
From the Zaak detail page, there needs to be a way to relate the viewed zaak with other zaken.

Parts that need to be added:

- [x]  The front end lets the user specify the `identificatie` and the `aard_relatie`. When the user types the `identificatie`, the potential zaken with that `identificatie` (for which the user has permissions) should be suggested. 

- [x]  The back end checks that the urls of the main zaak and the related zaak are not the same. Then, the relation is created. This can be similar to the code in BPTL: https://github.com/GemeenteUtrecht/bptl/blob/master/src/bptl/work_units/zgw/tasks/zaak_relations.py#L247

PR #114, opened 2020-11-09
109 - Create Zaken relations from zaak-detail page

SilviaAmAm:
Related ticket: #109 

An additional form in the zaak detail page was added. In this form, the related zaak and the type of relation can be filled out.

### DIFF
--- a/src/zac/core/api/permissions.py
+++ b/src/zac/core/api/permissions.py
@@ -45,13 +45,12 @@ class CanAddRelations(permissions.BasePermission):
         if not serializer.is_valid():
             return True
 
-        # Check that the user has access to both zaken being related
+        # Check that the user has access to the main zaak
         try:
-            get_zaak(zaak_url=serializer.validated_data["main_zaak"])
-            get_zaak(zaak_url=serializer.validated_data["relation_zaak"])
+            main_zaak = get_zaak(zaak_url=serializer.validated_data["main_zaak"])
         except ClientError:
             logger.info("Invalid Zaak specified", exc_info=True)
             return False
 
-        # Check that the user has permissions to add relations
-        return request.user.has_perm(zaken_add_relations.name)
+        # Check that the user has permissions to add relations to the main zaak
+        return request.user.has_perm(zaken_add_relations.name, main_zaak)
--- a/src/zac/core/api/permissions.py
+++ b/src/zac/core/api/permissions.py
@@ -5,7 +5,7 @@ from rest_framework.request import Request
 from rest_framework.views import APIView
 from zds_client import ClientError
 
-from ..permissions import zaken_add_documents
+from ..permissions import zaken_add_documents, zaken_add_relations
 from ..services import get_zaak
 
 logger = logging.getLogger(__name__)
@@ -33,3 +33,25 @@ class CanAddDocuments(permissions.BasePermission):
 
         zaak_url = serializer.validated_data["zaak"]
         return self._has_zaak_permission(request, zaak_url)
+
+
+class CanAddRelations(permissions.BasePermission):
+    def has_permission(self, request: Request, view: APIView) -> bool:
+        if request.method != "POST":
+            return False
+
+        serializer = view.get_serializer(data=request.data)
+        # if the serializer is not valid, we want to see validation errors -> permission is granted
+        if not serializer.is_valid():
+            return True
+
+        # Check that the user has access to both zaken being related
+        try:
+            get_zaak(zaak_url=serializer.validated_data["main_zaak"])
+            get_zaak(zaak_url=serializer.validated_data["relation_zaak"])
+        except ClientError:
+            logger.info("Invalid Zaak specified", exc_info=True)
+            return False
+
+        # Check that the user has permissions to add relations
+        return request.user.has_perm(zaken_add_relations.name)
--- a/src/zac/core/api/serializers.py
+++ b/src/zac/core/api/serializers.py
@@ -1,7 +1,9 @@
 from django.core.validators import RegexValidator
 from django.template.defaultfilters import filesizeformat
 from django.urls import reverse
+from django.utils.translation import gettext as _
 
+from djchoices import ChoiceItem, DjangoChoices
 from rest_framework import serializers
 
 from .utils import (
@@ -109,3 +111,15 @@ class ExtraInfoSubjectSerializer(serializers.Serializer):
     kinderen = serializers.ListField()
     verblijfplaats = serializers.DictField()
     partners = serializers.ListField()
+
+
+class AardRelatieChoices(DjangoChoices):
+    vervolg = ChoiceItem("vervolg", _("Vervolg"))
+    bijdrage = ChoiceItem("bijdrage", _("Bijdrage"))
+    onderwerp = ChoiceItem("onderwerp", _("Onderwerp"))
+
+
+class AddZaakRelationSerializer(serializers.Serializer):
+    relation_zaak = serializers.URLField(required=True)
+    aard_relatie = serializers.ChoiceField(required=True, choices=AardRelatieChoices)
+    main_zaak = serializers.URLField(required=True)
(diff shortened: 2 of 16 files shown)


src/zac/core/api/urls.py
```diff
@@ -19,6 +21,8 @@
         name="add-document",
     ),
     path("documents/info", GetDocumentInfoView.as_view(), name="get-document-info"),
+    path("zaken_relation", AddZaakRelationView.as_view(), name="add-zaak-relation"),
+    path("zaken", GetZakenView.as_view(), name="zaken"),
```
sergei-maertens, line 25: This URL name makes it seems like it's a frontend URL, probably that can be improved.

src/zac/core/api/views.py
```diff
@@ -149,3 +152,64 @@ def post(self, request: Request, **kwargs) -> Response:
...
+    def get_serializer(self, *args, **kwargs):
+        return AddZaakRelationSerializer(data=self.request.data)
+
+    def post(self, request: Request) -> Response:
+        serializer = self.get_serializer()
+        serializer.is_valid(raise_exception=True)
+
+        # Retrieving the main zaak
+        service = Service.get_service(serializer.validated_data["main_zaak"])
+        client = service.build_client()
```
sergei-maertens, line 169:
```suggestion
        client = Service.get_client(serializer.validated_data["main_zaak"])
```

src/zac/core/api/views.py
```diff
@@ -149,3 +152,64 @@ def post(self, request: Request, **kwargs) -> Response:
 
         serializer = ExtraInfoSubjectSerializer(extra_info_inp)
         return Response(serializer.data)
+
+
+class AddZaakRelationView(views.APIView):
+    permission_classes = (permissions.IsAuthenticated,)
```
sergei-maertens, line 158: this is not sufficient, it means I can now just post whatever to this endpoint as long as I'm logged in. There should be a permission check that you can add related zaken - the permission will have to be defined in https://github.com/GemeenteUtrecht/zaakafhandelcomponent/blob/develop/src/zac/core/permissions.py

src/zac/core/api/views.py
```diff
@@ -149,3 +152,64 @@ def post(self, request: Request, **kwargs) -> Response:
...
+        )
+
+        return Response(status=status.HTTP_200_OK)
+
+
+class GetZakenView(views.APIView):
+    permission_classes = (permissions.IsAuthenticated,)
+
+    def get_serializer(self, *args, **kwargs):
+        return ZaakIdentificatieSerializer(data=self.request.query_params)
```
sergei-maertens, line 193: you can put this directly in the `def get` - no need for a separate method

src/zac/core/api/views.py
```diff
@@ -149,3 +152,64 @@ def post(self, request: Request, **kwargs) -> Response:
...
+        for zaak in zaken:
+            data.append(
+                {
+                    "identificatie": serializer.validated_data["identificatie"],
+                    "bronorganisatie": zaak.bronorganisatie,
+                    "url": zaak.url,
+                }
+            )
+
+        return Response(data=data)
```
sergei-maertens, line 215: a serializer for this would make more sense and keep the view itself clean:

```py
class ZaakSerializer(serializers.Serializer):
    identificatie = serializers.CharField()
    bronorganisatie = serializers.CharField()
    url = serializers.URLField()

# view:
serializer = ZaakSerializer(instance=zaken, many=True)
```

src/zac/core/services.py
```diff
@@ -346,11 +346,18 @@ def get_zaken_es(
...
+
+    zrcs = Service.objects.filter(api_type=APITypes.zrc)
+    clients = [zrc.build_client() for zrc in zrcs]
 
     with parallel(max_workers=10) as executor:
-        results = executor.map(_get_zaak, zaak_urls)
+        job_args = []
+        for client, related_objects in zip(clients, zaak_urls):
+            job_args += [(client, zaak_url) for zaak_url in zaak_urls]
+        results = executor.map(_get_zaak, job_args)
```
sergei-maertens, line 360: You can _always_ derive which service/client you need from the zaak URL. This is amplifying the calls if there are multiple Zaken API services defined, because it will attempt to fetch the zaak in the other services and 404 there.

The original implementation was correct.

src/zac/js/views/zaak-detail/AddZaakRelation.js
```diff
@@ -0,0 +1,173 @@
...
+                </a>
+            </div>
+
+            <Modal
+              isOpen={ state.isModalOpen }
+              className='modal'
+              onRequestClose={closeModal}
+            >
+                <button onClick={ closeModal } className='modal__close btn'>&times;</button>
+                <h1 className='page-title'>Relatie toevoegen</h1>
```
sergei-maertens, line 127: `h1` is not correct for page accessibility. Maybe a `span` or `div` are more appropriate.

src/zac/js/views/zaak-detail/index.js
```diff
@@ -65,10 +70,45 @@ const initBetrokkenenTable = () => {
...
+        const data = new FormData();
+        data.append('csrfmiddlewaretoken', csrftoken);
+
+        await apiCall(
+            ENDPOINT_FLUSH_CACHE,
+            {
+                method: 'POST',
+                body: data,
+            }
+        );
```
sergei-maertens, line 94: There is no need for this - cache invalidation happens automatically with webhook callbacks. During development however you can't receive these callbacks, so you have to flush manually in that case.

sergei-maertens (changes requested):

sergei-maertens:
Functionality wise this is :+1:

src/zac/js/views/zaak-detail/index.js
sergei-maertens: There's one race-condition with the webhook call - you could as part of the endpoint to relate the zaak already invalidate the detail cache yourself - see `zac.core.cache.invalidate_zaak_cache`

src/zac/core/api/serializers.py
```diff
@@ -109,3 +111,28 @@ class ExtraInfoSubjectSerializer(serializers.Serializer):
     kinderen = serializers.ListField()
     verblijfplaats = serializers.DictField()
     partners = serializers.ListField()
+
+
+class AardRelatieChoices(DjangoChoices):
```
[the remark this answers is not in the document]
SilviaAmAm: Created PR: https://github.com/maykinmedia/zgw-consumers/pull/15

src/zac/core/api/urls.py
SilviaAmAm: Changed it to `zaken_search`

src/zac/core/api/permissions.py
```diff
@@ -33,3 +33,25 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+        # Check that the user has access to both zaken being related
+        try:
+            get_zaak(zaak_url=serializer.validated_data["main_zaak"])
+            get_zaak(zaak_url=serializer.validated_data["relation_zaak"])
+        except ClientError:
+            logger.info("Invalid Zaak specified", exc_info=True)
+            return False
+
+        # Check that the user has permissions to add relations
+        return request.user.has_perm(zaken_add_relations.name)
```
sergei-maertens, line 57: This does not do what you think it does - the `get_zaak` function only looks it up. It does check that the ZAC itself has access to both zaken.

However, what should be checked is that the user has permission to add a relation to the main zaak. I wouldn't enforce that they must have permission to view the related zaak (at first).

This would turn the permission into:

```py
zaak = get_zaak(zaak_url=serializer.validated_data["main_zaak"])
return request.user.has_perm(zaken_add_relations.name, zaak)
```

src/zac/core/permissions.py
```diff
@@ -45,3 +45,8 @@
 zaken_request_access = Permission(
     name="zaken:toegang-aanvragen", description="Toegang aanvragen voor zaken"
 )
+
+zaken_add_relations = Permission(
+    name="zaken:nieuwe-relaties-toevoegen",
+    description="Voegen relaties tussen zaken toe.",
```
sergei-maertens, line 51:
```suggestion
    description="Relateer andere zaken aan de (hoofd)zaak",
```

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #124, opened 2020-11-17
Fixing race condition bug on Kownsl approval/advice

sergei-maertens:
When Kownsl advices/approvals are recorded, notifications are sent and
handled by the ZAC. These handlers comlete the user task in the
Camunda process. Sometimes, these notifications can arrive before the
redirect is triggered, which also completes the user task.

The code is now adapted for this race condition where the notification
is handled before the redirect. In that case, the return URL validates
against the historic process instance/task, and if it's a historic
entry, it no longer attempts to complete the user task (since that
has been done already)

- [x] Reproduce
- [x] Handle race condition
- [x] Tests for history API fallback

### DIFF
--- a/src/zac/camunda/history.py
+++ b/src/zac/camunda/history.py
@@ -1,0 +1,20 @@
+from typing import Any
+
+from django_camunda.client import get_client
+from django_camunda.types import CamundaId
+from django_camunda.utils import deserialize_variable
+
+
+def get_historical_variable(instance_id: CamundaId, name: str) -> Any:
+    client = get_client()
+
+    response = client.get(
+        "history/variable-instance",
+        {
+            "processInstanceId": instance_id,
+            "variableName": name,
+            "deserializeValues": "false",
+        },
+    )
+    values = sorted(response, key=lambda val: val["create_time"], reverse=True)
+    return deserialize_variable(values[0])
--- a/src/zac/core/camunda.py
+++ b/src/zac/core/camunda.py
@@ -56,12 +56,15 @@ def get_process_tasks(process: ProcessInstance) -> List[Task]:
     return tasks
 
 
-def get_task(task_id: CamundaId) -> Optional[Task]:
+def get_task(task_id: CamundaId, check_history=False) -> Optional[Task]:
     client = get_client()
     try:
         data = client.get(f"task/{task_id}")
     except requests.HTTPError as exc:
         if exc.response.status_code == 404:
+            if not check_history:
+                return None
+
             # see if we can get it from the history
             historical = client.get("history/task", {"taskId": task_id})
             if not historical:
--- a/src/zac/core/camunda.py
+++ b/src/zac/core/camunda.py
@@ -3,9 +3,11 @@ from typing import List, Optional
 from django.contrib.auth import get_user_model
 
 import requests
-from django_camunda.camunda_models import factory
 from django_camunda.client import get_client
 from django_camunda.types import CamundaId
+from zgw_consumers.api_models.base import (  # django_camunda can't handle inheritance yet
+    factory,
+)
 
 from zac.camunda.data import ProcessInstance, Task
 from zac.camunda.forms import extract_task_form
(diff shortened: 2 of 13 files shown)


src/zac/camunda/history.py
```diff
@@ -0,0 +1,20 @@
...
+    response = client.get(
+        "history/variable-instance",
+        {
+            "processInstanceId": instance_id,
+            "variableName": name,
+            "deserializeValues": "false",
+        },
+    )
+    values = sorted(response, key=lambda val: val["create_time"], reverse=True)
+    return deserialize_variable(values[0])
```
sergei-maertens, line 20: I have no idea how this looks like when variables change throughout the process - with the current testing there was only a single historical entry for a variable.

src/zac/core/views/processes.py
```diff
@@ -364,8 +364,9 @@ def validate_state(self, encoded_state: str) -> dict:
         return state
 
     def complete_task(self) -> HttpResponseRedirect:
-        task = self._get_task()
-        complete_task(task.id, {})
+        task = self._get_task(refresh=True)
+        if not task.historical:
+            complete_task(task.id, {})
```
sergei-maertens, line 369: This is the actual fix - by refreshing as close to the `complete_task` call, we minimize the risk of race conditions. It does not completely eliminate them, but I'd rather see how far we get without wrapping this in all kinds of defensive programming constructions.

If we know that the task is fetched from the history, this implies that it was completed already (from the notifications, in this case).

src/zac/camunda/history.py
damm89: Empty lists won't be in the response/aren't relevant?

In the camunda docs it only says it'll return a 400 if query_params is invalid. Else response is always 200?

src/zac/core/camunda.py
```diff
@@ -62,8 +62,30 @@ def get_task(task_id: CamundaId) -> Optional[Task]:
...
-            return None
-        raise
+            # see if we can get it from the history
+            historical = client.get("history/task", {"taskId": task_id})
+            if not historical:
+                return None
+
+            assert (
+                len(historical) < 2
+            ), f"Found multiple tasks in the history for ID {task_id}"
```
damm89, line 72: I don't understand why this should raise an assertion error.

src/zac/core/views/processes.py
damm89: Why not silently catch the exception here if it's already completed and redirect to a different page saying the task was already completed?

damm89 (commented):
Have some questions!

src/zac/camunda/history.py
sergei-maertens: yeah I trust the Camunda docs. Empty list is technically possible I suppose if you use a variable name that doesn't exist, but in this case `zaakUrl` is pretty much always set. If we start using this function in a broader sense, we'll have to check if there are edge cases relevant. I wanted to prevent blowing up the PR even more.

src/zac/core/camunda.py
sergei-maertens: The API is weird for me, I don't understand how there's not a detail endpoint for it. The task IDs are UUIDs, so they should be unique. Since we're doing the lookup based on ID, we really expect a single result to be returned - especially since the lookup should be initiated by something that originated from a non-history task. If you're just guessing IDs, then this will 404 for the client.

It's basically a sanity check expressing that we never expect more than one result, and if we do, all systems should yell very loudly to figure out what assumption is wrong.

src/zac/core/views/processes.py
sergei-maertens: Because it's not a user mistake - for the user this is silent. The task was already completed in another way, and they just got automatically redirected to this page, which redirects the user back to the detail page. This is not an explicit user action where the user clicks the button "complete task" - it's entirely automated.

So, I want to behaviour to be the same independent from what happens first: either the notification is handled first, or the user redirect is handled first (which corresponds by going to Kownsl from e-mail vs. going to Kownsl from the ZAC tasks)

src/zac/core/camunda.py
```diff
@@ -62,8 +62,30 @@ def get_task(task_id: CamundaId) -> Optional[Task]:
         data = client.get(f"task/{task_id}")
     except requests.HTTPError as exc:
         if exc.response.status_code == 404:
-            return None
-        raise
+            # see if we can get it from the history
+            historical = client.get("history/task", {"taskId": task_id})
```
annashamray, line 66: I'm not sure that retrieving historical data in a function which is widely used in the app is a good idea, sometime in the future we can face the situation when the data is shown for the task which is already closed. I've checked the current use of `get_zaak` and it should be safe, but we always need to keep it in mind now

Perhaps adding a bool `use_historic` to this function could ease our future development

sergei-maertens: I have played with that idea too, seems like a safe option to add it indeed! Will do

sergei-maertens: Fixed.

annashamray (approved):

damm89 (approved):

PR #131, opened 2020-12-08
Add simple support for exports/reports

sergei-maertens:
This feature adds support for simple tabular data reports, per https://github.com/GemeenteUtrecht/ZGW/issues/638

* you can define reports in the admin. You give them a name, and a list of zaaktype identificaties
* the reports views provide:
    * list view to view the reports available to you - a report is available if you have access to _all_ the zaaktypen in the report (based on the auth profiles)
    * download view to generate the export and serve it as XLSX file (download)
* the export itself uses `tablib`, with currently hard-coded headers. Status, eigenschappen and some other zaak-attributes are used. The elasticsearch index is used to filter the zaken on zaaktype and only the not-closed zaken.


src/zac/reports/export.py
```diff
@@ -0,0 +1,101 @@
...
+def _get_zaaktypen(report: Report) -> List[ZaakType]:
+    identificaties = report.zaaktypen
+
+    def _get_zaaktype_versions(identificatie: str):
+        return _get_from_catalogus(
+            "zaaktype", catalogus="", identificatie=identificatie
+        )
+
+    with parallel() as executor:
+        zaaktypen = executor.map(_get_zaaktype_versions, identificaties)
```
damm89, line 29: why not something like:

```suggestion
    with parallel() as executor:
        zaaktypen = executor.map(
            lambda identificatie: _get_from_catalogus(
                "zaaktype", catalogus="", identificatie=identificatie
            ),
            identificaties,
        )
```
or does it mess with readability too much? :-)

src/zac/reports/export.py
```diff
@@ -0,0 +1,101 @@
...
+    zaken = get_export_zaken(report)
+
+    # get the statuses & eigenschappen
+    with parallel() as executor:
+        statuses: Iterator[Status] = executor.map(get_status, zaken)
+        eigenschappen: Iterator[List[ZaakEigenschap]] = executor.map(
+            get_zaak_eigenschappen, zaken
+        )
+
+    zaak_statuses: Dict[str, str] = {
```
damm89, line 63: This kind of variable typing is new to me. Is that a new pep guideline?

damm89 (approved):

src/zac/reports/export.py
```diff
@@ -0,0 +1,101 @@
...
+
+    data = tablib.Dataset(
+        headers=[
+            "zaaknummer",
+            "zaaktype",
+            "startdatum",
+            "omschrijving",
+            "eigenschappen",
+            "status",
+        ]
```
annashamray, line 80: Despite Michelle hasn't asked for this but I think it'd be a good idea to add `bronorganisatie` to the report - then we will be able to construct ZAC url to the detail page ot the reported zaken.

annashamray (commented):
This feature will definetely help to win excel fans' hearts :) 

But I don't think that spliting the user experience between admin page (to create report object) and the website is the best approach. 
Would it be better to filter zaaktypen on the report page? Or maybe just add report/download button to zaak-list page? 
But we can allways do it in the next iterations

src/zac/reports/export.py
sergei-maertens: https://www.python.org/dev/peps/pep-0484/ probably says something about it

src/zac/reports/export.py
sergei-maertens: There's not really a need for this - the exports will just serve to discuss things with people without access to the systems, so there's no point in having the URL either. People with access can also use the search functionality to find the zaken.

Splitting admin page config vs. frontend was discussed last friday - this is purely not worth the investment at this time considering the SPA that's in development. Dynamic filtering is not desired, because they are the same reports/configurations that should be exported periodically - pre-defining them saves the people time and keeps reports consistent.

A generic export on zaak-list could be a nice future feature.

PR #133, opened 2020-12-11
Added functionality to get path to UI for backend

damm89:
It works - but doesn't feel clean. @sergei-maertens ?


sergei-maertens (approved):
we're dealing with URLs that are outside of Django's scope, don't think there's a cleaner way.

sergei-maertens:
Failing tests are not caused by this PR

PR #139, opened 2020-12-23
Invalidate cache on mutation of zaaktype/informatieobjecttype

sergei-maertens:
Fixes a `KeyError` when retrieving a zaak of a newly created zaaktype, while the cache of zaaktypen is still warm and doesn't include the new zaaktype yet.

### DIFF
--- a/backend/src/zac/notifications/handlers.py
+++ b/backend/src/zac/notifications/handlers.py
@@ -1,0 +1,82 @@
+from zgw_consumers.api_models.base import factory
+from zgw_consumers.api_models.zaken import Zaak
+
+from zac.activities.models import Activity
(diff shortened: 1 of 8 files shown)


backend/src/zac/notifications/handlers.py
```diff
@@ -0,0 +1,107 @@
...
+                    catalogus=data["kenmerken"]["catalogus"]
+                )
+
+
+class RoutingHandler:
+    def __init__(self, config: dict, default=None):
+        self.config = config
+        self.default = default
+
+    def handle(self, message: dict):
```
damm89, line 93:
```suggestion
    def handle(self, message: dict) -> None:
```

annashamray (approved):
Nice refactoring of notification handlers!

damm89 (commented):
Minor comment so far.

PR #149, opened 2021-01-08
Added endpoints to fetch review request data from kownsl for BFF

damm89:


backend/src/zac/core/api/bff_urls.py
```diff
@@ -46,4 +48,14 @@
...
+    path(
+        "cases/<str:bronorganisatie>/<str:identificatie>/review-requests/completed",
+        ZaakReviewRequestCompletedView.as_view(),
+        name="zaak-review-requests-completed",
+    ),
+    path(
+        "cases/<str:bronorganisatie>/<str:identificatie>/review-requests/detail",
+        ZaakReviewRequestDetailView.as_view(),
+        name="zaak-review-requests-detail",
+    ),
```
sergei-maertens, line 60: I think we should keep Kownsl-related things in `zac.contrib.kownsl`, so I'd move these URLs and views to that package. I'd also re-structure the URLs a bit for that:

```py
# in zac.contrib.kownsl.urls

urlpatterns += [
    path(
        "zaak-review-requests/<str:bronorganisatie>/<str:identificatie>",
        ZaakReviewRequestsView.as_view(),
        name="zaak-review-requests",
    )
]
```

Also notice that I'm condensing both endpoints into one here - see later remarks.

backend/src/zac/core/api/serializers.py
```diff
@@ -391,3 +398,76 @@ class ZaakObjectGroupSerializer(APIModelSerializer):
...
+    completed = serializers.SerializerMethodField(
+        label=_("completed requests"), help_text=_("The number of completed requests.")
+    )
+
+    class Meta:
+        model = ReviewRequest
+        fields = ("id", "review_type", "completed", "num_assigned_users")
+
+    def get_completed(self, obj) -> int:
+        return obj.num_advices + obj.num_approvals
```
sergei-maertens, line 413: The serializer name gave me the impression that this would only have the `completed` field, but there's more information of course.

backend/src/zac/core/api/serializers.py
```diff
@@ -391,3 +398,76 @@ class ZaakObjectGroupSerializer(APIModelSerializer):
...
+
+    class Meta:
+        model = ReviewRequest
+        fields = ("id", "review_type", "reviews")
+
+    def get_reviews(self, obj) -> Optional[List[dict]]:
+        if obj.review_type == "advice":
+            return AdviceSerializer(obj.advices, many=True).data
+        else:
+            return ApprovalSerializer(obj.approvals, many=True).data
```
sergei-maertens, line 473: We have a case of polymorphism going on here, and I'd like that to be reflected in the API documentation. I'll have a go at this on this branch.

backend/src/zac/core/api/views.py
```diff
@@ -293,6 +300,16 @@ def get(self, request, *args, **kwargs):
...
 
         review_requests = get_review_requests(zaak)
+
+        with parallel() as executor:
+            _advices = executor.map(
+                lambda rr: retrieve_advices(rr) if rr else [],
+                [rr if rr.num_advices else None for rr in review_requests],
+            )
+
+            for rr, rr_advices, rr_approvals in zip(review_requests, _advices):
```
sergei-maertens, line 310: `rr_approvals` is not defined, that `zip` tuple will not match so I don't expect this to work at all.

backend/src/zac/core/api/views.py
```diff
@@ -361,3 +378,43 @@ def group_key(zo):
...
+class ZaakReviewRequestCompletedView(GetZaakMixin, views.APIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAuthenticated & CanReadZaken,)
+    serializer_class = ZaakRevReqCompletedSerializer
+
+    def get(self, request, *args, **kwargs):
+        zaak = self.get_object()
+        review_requests = get_review_requests(zaak)
+        serializer = self.serializer_class(instance=review_requests, many=True)
+        return Response(serializer.data)
```
sergei-maertens, line 392: Please move this to `zac.contrib.kownsl`.

I'd also rename this as `ZaakReviewRequestsSummaryView` - as in the UI it displays both the type, number completed and number requested.

backend/src/zac/core/api/views.py
```diff
@@ -361,3 +378,43 @@ def group_key(zo):
...
+            )
+
+            for rr, rr_advices, rr_approvals in zip(
+                review_requests, _advices, _approvals
+            ):
+                rr.advices = rr_advices
+                rr.approvals = rr_approvals
+
+        serializer = self.serializer_class(instance=review_requests, many=True)
+        return Response(serializer.data)
```
sergei-maertens, line 420: Looking at the UI (current), it doesn't make sense to me to fetch all the review request data in bulk. I.e., a zaak can have multiple review requests, but if you're bringing up the detail information, you're doing it for a single one at once.

1. the URL should take an `id` or `uuid` parameter (`bronorganisatie` and `identificatie` are then not required), and that particular review request should then be retrieved from Kownsl.
2. Once that is done, you can retrieve `review_request.for_zaak` and perform the permission checks on that zaak (`GetZaakMixin` behaviour)
3. Once everything checks out, you can serialize just that single RR and not have to deal with any overhead of the review requests that don't have to be displayed (yet). I presume that will also simplify the frontend code.

sergei-maertens (changes requested):

backend/src/zac/core/api/views.py
sergei-maertens: Come to think of it... do we even need this endpoint? What does it offer that http://localhost:8000/api/docs/#operation/kownsl_review_requests_advice_retrieve doesn't?

damm89: Still no need to pull all review requests, you're right. :-)

backend/src/zac/core/api/views.py
damm89: Yep, done.

backend/src/zac/core/api/serializers.py
damm89: Curious as to see your solution!

backend/src/zac/contrib/kownsl/api.py
```diff
@@ -83,3 +85,23 @@ def get_review_requests(zaak: Zaak) -> List[ReviewRequest]:
...
+    # operation_id in the schema (yet?). Building the url and doing the request
+    # manually for now.
+    try:
+        url = client.get_operation_url(
+            client.schema, "reviewrequest_retrieve", **{"uuid": uuid}
+        )
+        result = client.request(url)
+
+    except ClientError:
+        return Http404(f"Review request with id {uuid} does not exist.")
```
sergei-maertens, line 104:
```suggestion
        raise Http404(f"Review request with id {uuid} does not exist.")
```

backend/src/zac/contrib/kownsl/serializers.py
```diff
@@ -0,0 +1,87 @@
...
+        fields = ("id", "review_type", "completed", "num_assigned_users")
+
+    def get_completed(self, obj) -> int:
+        return obj.num_advices + obj.num_approvals
+
+
+class AuthorSerializer(APIModelSerializer):
+    class Meta:
+        model = Author
+        fields = ("first_name", "last_name")
```
sergei-maertens, line 33: I'd include `username` too, because `first_name` and `last_name` can be blank, so the frontend can at least fall back to the username in that case.

backend/src/zac/contrib/kownsl/views.py
```diff
@@ -134,3 +147,36 @@ class AdviceRequestView(BaseRequestView):
...
+
+        try:
+            zaak = get_zaak(review_request.for_zaak)
+        except ObjectDoesNotExist:
+            raise Http404(f"No zaak is found for url: {review_request.for_zaak}.")
+
+        self.check_object_permissions(self.request, zaak)
+
+        review_request.advices = retrieve_advices(review_request)
+        review_request.approvals = retrieve_approvals(review_request)
```
sergei-maertens, line 180: Small :zap: performance improvement possible by doing this in a `parallel()` executor

sergei-maertens (changes requested):
Looks good, couple more minor remarks to put the icing on the cake

backend/src/zac/core/api/serializers.py
sergei-maertens: Hope to get to it today... but it's not the most urgent thing so not a blocker for this PR.

sergei-maertens (approved):

PR #157, opened 2021-01-21
Feature/document wijzigen

damm89:
The client should be able to receive input, process it and request information from the D.O.C. (rebranding to dowc).

If a client tries to POST a duplicate request, the dowc returns a 409 conflict error which the ZDSClient will catch in its assertion of the create method. It will then retry a list request with filters to fetch the requested data from the dowc.

@kelvincy is fixing the buttons and responses on the frontend of the zaak-detail page.


backend/src/zac/core/fields.py
```diff
@@ -12,7 +12,7 @@ def __init__(self, doc):
     @property
     def download_url(self):
         download_path = reverse(
-            "core:download-document",
+            "dowc:download-document",
```
damm89, line 15:
```suggestion
            "core:download-document",
```

damm89 (commented):
Fixed an error.

backend/src/zac/core/api/serializers.py
```diff
@@ -299,10 +307,16 @@ class Meta:
...
-        label=_("ZAC download URL"),
+    read_url = serializers.SerializerMethodField(
+        label=_("ZAC document read URL"),
         help_text=_(
-            "The download URL for the end user. Will serve the file as attachment."
+            "The document URL for the end user that opens a document to be read. Will serve the file from a WebDAV server."
+        ),
+    )
+    write_url = serializers.SerializerMethodField(
+        label=_("ZAC document read URL"),
```
damm89, line 317:
```suggestion
        label=_("ZAC document write URL"),
```

backend/src/zac/contrib/dowc/views.py
```diff
@@ -0,0 +1,59 @@
...
+    return type_(value)
+
+
+class OpenDowcView(APIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAuthenticated & CanOpenDocuments,)
+    document = None
+    serializer_class = DowcResponseSerializer
+
+    def get_object(self) -> NoReturn:
```
sergei-maertens, line 28:
```suggestion
    def get_object(self) -> Document:
```
with `zgw_consumers.api_models.documenten.Document`

backend/src/zac/contrib/dowc/views.py
```diff
@@ -0,0 +1,59 @@
...
+            {
+                "magic_url": document.url,
+                "delete_url": document.url,
+                "purpose": purpose,
+            }
+        )
+        drc_url = self.document.url
+        dowc_response, status_code = get_doc_info(request.user, drc_url, purpose)
+        serializer = self.serializer_class(dowc_response)
+        return Response(serializer.data, status=status_code)
```
sergei-maertens, line 50: second return statement will never hit

sergei-maertens (changes requested):

backend/src/zac/contrib/dowc/views.py
damm89: Yep, that was the mock data mentioned in slack. :-)

sergei-maertens (approved):

PR #163, opened 2021-01-28
Set up basic architecture of user-task data + context

sergei-maertens:
* Endpoint defined and documented in ReDoc taking authenticated requests with a task ID URL parameter
* Basic plumbing to provide mapping between `Task.formKey` and specific callback to retrieve the relevant context
* implement Polymorphism (and OpenAPI documentation of that). Again.

TODO (in other PRs):

* hook up the API documentation with polymorphism
* implement functional context functions + matching serializers
* Tests


damm89 (commented):
Looking good for a first iteration. We'll write tests after demo?

sergei-maertens:
> Looking good for a first iteration. We'll write tests after demo?

yeah, too much work to be able to deliver something functional

backend/src/zac/camunda/api/__init__.py
```diff
@@ -1,11 +1,14 @@
...
 urlpatterns = [
     path(
         "fetch-process-instances",
         ProcessInstanceFetchView.as_view(),
         name="fetch-process-instances",
     ),
+    path(
+        "task-data/<uuid:task_id>", GetTaskContextView.as_view(), name="get-task-data"
+    ),
 ]
```
damm89, line 14: why not in urls.py?

sergei-maertens: no particular reason, I don't remember, so I'll move it :-)

backend/src/zac/contrib/kownsl/camunda.py
```diff
@@ -0,0 +1,35 @@
...
+class AdviceApprovalContextSerializer(APIModelSerializer):
+    class Meta:
+        model = AdviceApprovalContext
+        fields = ("review_type", "title")
+
+
+@register("zac:configureApprovalRequest", AdviceApprovalContextSerializer)
+@register("zac:configureAdviceRequest", AdviceApprovalContextSerializer)
+def get_context(task: Task) -> AdviceApprovalContext:
+    return AdviceApprovalContext(
```
damm89, line 30: Not sure what's happening here.

backend/src/zac/camunda/user_tasks/context.py
```diff
@@ -0,0 +1,75 @@
...
+
+    return decorator
+
+
+EmptySerializer = usertask_context_serializer(serializers.JSONField)
+
+
+@register("", EmptySerializer)
+@register("zac:documentSelectie", EmptySerializer)
+@register("zac:gebruikerSelectie", EmptySerializer)
```
damm89, line 73: Just leave these guys out?

damm89 (commented):
Hard to follow what's going on. Looks impressive though.

backend/src/zac/contrib/kownsl/camunda.py
sergei-maertens: There are essentially two essential parts to a particular `formKey` context:

* a getter function that can construct the context, given the Camunda Task. For example, look up some task/process variables, look up related data to those variables from other systems (zaken API, Kownsl... etc.)
* a serializer to serialize that context

The third part is a domain-specific object, in this case the `AdviceApprovalContext`. It is the object that is fed to the serializer, as the involved classes (Zaak, ZaakType etc.) likely have more properties that are not relevant for the context. You build the context as Python datastructures, and feed them to the serializer to turn them into JSON.

The `@register` decorator registers the Camunda task `formKey` with the given serializer and getter function. So, when the task data details are retrieved, the following happens:

1. the task is fetched from Camunda 
2. the `formKey` is looked up in the registry, revealing the appropriate getter function
3. the getter function is called with the task, which returns a particular Context
4. the task data, `formKey` value and context are passed to the polymorphic `TaskData` serializer, which serializes everything

The code you comment on, is the mechanism for each (contrib) app to register their own `formKeys`. In this case, I've set up a static, hardcoded context without any lookups to remove API's purely for mechanism demonstration/testing purposes.

backend/src/zac/camunda/user_tasks/context.py
sergei-maertens: Could do, but I wanted to include example code on how to register form keys :) Can be removed when we have actual, real implementations.

backend/src/zac/contrib/kownsl/camunda.py
damm89: Thanks for the explanation.

PR #164, opened 2021-01-28
Feature/es search endpoint

### LINKED ISSUE #147: Implement endpoint to perform search
We need an endpoint for the SPA in the BFF to perform the search exposed to the end-user, see https://github.com/GemeenteUtrecht/ZGW/issues/605

Given the complexity of the search form, this will have to be a `POST` action with a request body.

```http
POST /api/v1/core/zaken/search HTTP/1.1
Content-Type: application/json
Cookie: zac-sessionid=...
Accept: application/json

{
  "identificatie": "abcd",
  "zaaktype": {
    "omschrijving": "zaaktype 1",
    "catalogus": "https://openzaak.utrechtproeftuin.nl/catalogi/api/v1/..."
  },
  "omschrijving": "foo bar baz",
  "eigenschappen": {
    "Buurt": {
      "value": "Leidsche Rijn"
    }
  }
}
```

* `identificatie` performs a case-insensitive "contains" search. Could be full-text-search in the future.
* `zaaktype` initially provides case-insensitive search on zaaktype omschrijving. This means that all versions of that zaaktype must be considered. To resolve these, you need the `catalogus` which the frontend obtains from #145 endpoints.
* `omschrijving` should perform full-text search like on `zaak.omschrijving` text. Let's start with ES defaults for analyzers etc.
* `eigenschappen`: let's make this multi-valued in case the future demands to filter on multiple eigenschappen at the same time. The key is the name of the property (which is indexed in ES given #146), the value object contains a `value` key at first, but we leave room to expand this in the future with extra meta-data (such as `contains` / gte / etc.).

**Acceptance criteria**

- [x] All keys are optional
- [x] All keys are `AND`-ed together
- [x] The backend hits the ES index of zaken to obtain the zaak URLs for optimal performance
- [x] Response contains `zaak` with included `zaak.zaaktype` resolved (there are existing serializers)

annashamray:
fixes #147

based on #158


backend/src/zac/core/api/bff_urls.py
```diff
@@ -50,4 +51,5 @@
     ),
     path("zaaktypen", ZaakTypenView.as_view(), name="zaaktypen"),
     path("eigenschappen", EigenschappenView.as_view(), name="eigenschappen"),
+    path("search", SearchViewSet.as_view(), name="search"),
```
damm89, line 54: I'm wondering if this isn't better suited in the elasticsearch app rather than under core/api.

backend/src/zac/core/api/tests/test_search.py
```diff
@@ -0,0 +1,320 @@
...
+            "omschrijving": "some",
+        }
+
+        response = self.client.post(self.endpoint, data=data)
+
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        data = response.json()
+        self.assertEqual(len(data), 1)
+        self.assertEqual(data[0]["url"], zaak1["url"])
```
damm89, line 226: Maybe it's a silly remark and a bit unnecessary, but it would be nice to see what the response looks like.

backend/src/zac/core/api/tests/test_search.py
```diff
@@ -0,0 +1,320 @@
...
+        update_eigenschappen_in_zaak_document(zaak1_model)
+        update_eigenschappen_in_zaak_document(zaak2_model)
+        self.refresh_index()
+
+        data = {"eigenschappen": {"Buurt": {"value": "Leidsche Rijn"}}}
+
+        response = self.client.post(self.endpoint, data=data)
+
+        data = response.json()
+        self.assertEqual(len(data), 1)
```
damm89, line 319: See earlier remark on data object.

backend/src/zac/core/api/views.py
```diff
@@ -467,3 +467,34 @@ def get_eigenschappen(self, zaaktypen):
...
+        if data.get("zaaktype"):
+            zaaktype_data = data.pop("zaaktype")
+            zaaktypen = get_zaaktypen(
+                user_perms,
+                catalogus=zaaktype_data["catalogus"],
+                omschrijving=zaaktype_data["omschrijving"],
+            )
+            data["zaaktypen"] = [zaaktype.url for zaaktype in zaaktypen]
+
+        zaken = get_zaken_es(user_perms, size=50, query_params=data)
```
damm89, line 499: Again I'm wondering if this isn't better suited in elasticsearch app.

damm89 (commented):
Just some final remarks/questions, should've probably included them with the last review. This will be very nice once it's implemented. :+1: :+1:

backend/src/zac/core/api/bff_urls.py
sergei-maertens: Hrm, I'm not sure that we want to expose elasticsearch in the API URLs though, as in, not even hint at the underlying search engine.

backend/src/zac/core/api/views.py
sergei-maertens: `core` app deals with Zaken API and interaction of that, `elasticsearch` app should only offer the interfaces to hit the search index. I'm okay with this, but we can re-evaluate

sergei-maertens (approved):

backend/src/zac/core/api/views.py
annashamray: Since Sergei rearranged api endpoints in #174, I'll move this endpoint to `search` group

backend/src/zac/core/api/tests/test_search.py
annashamray: added assertion for the full response

backend/src/zac/core/api/tests/test_search.py
annashamray: added assertion for the full response

sergei-maertens (approved):

PR #166, opened 2021-01-30
Added a return to ui/zaken/<bronorganisatie>/<identificatie>/ url to …

damm89:
…get review request endpoint

### DIFF
--- a/backend/src/zac/contrib/kownsl/views.py
+++ b/backend/src/zac/contrib/kownsl/views.py
@@ -15,6 +15,7 @@ from zgw_consumers.concurrent import parallel
 from zgw_consumers.models import Service
 
 from zac.core.api.permissions import CanReadZaken
+from zac.core.api.serializers import ZaakSerializer
 from zac.core.api.views import GetZaakMixin
 from zac.core.services import get_zaak
 from zac.core.utils import get_ui_url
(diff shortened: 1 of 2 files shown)


backend/src/zac/contrib/kownsl/views.py
```diff
@@ -113,7 +114,18 @@ def get(self, request, request_uuid):
...
+        zaak = get_zaak(zaak_url)
+        return_url = get_ui_url(
+            [
+                "ui",
+                "zaken",
+                zaak.bronorganisatie,
+                zaak.identificatie,
+            ]
+        )
+        data = {**review_request, "return_url": return_url}
```
sergei-maertens, line 127: this is called `frontend_url` in the Kownsl schema, not `return_url`

```suggestion
        data["frontend_url"] = return_url
```

sergei-maertens (changes requested):
I don't understand where those frontend package changes keep coming from? Can you maybe just not commit those?

backend/src/zac/contrib/kownsl/views.py
damm89: Are you sure?

From https://kownsl.utrechtproeftuin.nl/api/docs/:
frontend_url: The frontend URL for reviewers to submit their review.

Looking at https://github.com/GemeenteUtrecht/kownsl/blob/master/src/kownsl/api/serializers.py this seems to refer back to "core:detail" of kownsl. 

This is to go return to the zaak-detail page.

sergei-maertens: in that case I am wrong - but I don't understand why the backend then has to provide that URL?

sergei-maertens (approved):

PR #178, opened 2021-02-03
Added dynamic form context to user task endpoint - api docs not worki…

damm89:
…ng yet.

Part of the dynamic forms will return a different value than the other ones. Think of "enum" input_type, but also data validation and expected data is different based on input_type. That could be handled internally without explicitly exposing that, of course, but that's not as neat. (see ZaakRevReqDetailSerializer for example) 

Not sure if I should continue down this path. I thought it'd be nice if all values get validated by their respective serializers and the API docs would kick back the appropriate data type instead of just generic data. 

@sergei-maertens:
Okay, got it to semi-work. On the redocs page you can't change the inputType yet. Would you have a clue? I'm guessing it has to do with the way the nested serializer is detailed in the schema.

### DIFF
--- a/backend/src/zac/camunda/dynamic_forms/context.py
+++ b/backend/src/zac/camunda/dynamic_forms/context.py
@@ -1,0 +1,70 @@
+from typing import List, Optional
+from xml.etree.ElementTree import Element
+
+from django_camunda.bpmn import CAMUNDA_NS
+
+from zac.core.camunda import get_process_zaak_url
+from zac.core.services import fetch_zaaktype, get_zaak
+
+from ..bpmn import get_bpmn
+from ..data import Task
+from ..process_instances import get_process_instance
+from ..user_tasks.context import register
+from .data import DynamicFormContext, FormField
+from .form_field import get_form_field_context
+from .serializers import DynamicFormContextSerializer
+
+
+def get_form_fields_from_task(task) -> Optional[List[Element]]:
+    """
+    Retrieve form fields from camunda task.
+    """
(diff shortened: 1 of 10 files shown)


backend/src/zac/camunda/dynamic_forms/context.py
```diff
@@ -0,0 +1,70 @@
...
+                name=definition.attrib["id"],
+                label=definition.attrib.get("label", ""),
+                input_type=input_type,
+                form_field_context=get_form_field_context(input_type, **kwargs),
+            )
+        )
+    return form_fields
+
+
+@register("zac:dynamicForm", DynamicFormContextSerializer)
```
damm89, line 52: @sergei-maertens : This `@register` doesn't actually register for some reason.

backend/src/zac/camunda/dynamic_forms/serializers.py
```diff
@@ -0,0 +1,126 @@
...
+}
+
+
+@register_input_type("zac:form_field:boolean", BooleanFieldSerializer)
+@register_input_type("zac:form_field:date", DateTimeFieldSerializer)
+@register_input_type("zac:form_field:long", IntFieldSerializer)
+@register_input_type("zac:form_field:string", StringFieldSerializer)
+@register_input_type("zac:form_field:enum", ChoiceFieldContextSerializer)
+def get_form_field_context(input_type: str, **kwargs):
+    return INPUT_TYPE_FIELD_MAPPING[input_type](**kwargs)
```
damm89, line 126: Neither do these.

Issue #175, opened 2021-02-03
ZAC adviseren/accorderen detail


kelvincy:
On calling `/api/kownsl/zaak-review-requests/7c384342-d28f-4a8a-94da-f67aef0c6991/detail`

I am receiving the following 500 error:
`AttributeError at /api/kownsl/zaak-review-requests/7c384342-d28f-4a8a-94da-f67aef0c6991/detail
'Client' object has no attribute 'get_operation_url'`

PR #179, opened 2021-02-04
Fixed get_review_request

damm89:
Fixes https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/175


sergei-maertens (approved):
Tentatively approving this - there is a cleaner way now in the latest gemma-zds-client version where you can configure the operation suffix map: https://pypi.org/project/gemma-zds-client/#resources-manipuleren

This should make it possible to just do `client.retrieve("reviewrequest", uuid=uuid)` instead

PR #180, opened 2021-02-04
Dynamic forms task data context

sergei-maertens:
Alternative implementation for #178 

- [x] Working POC
- [x] Tests

### DIFF
--- a/backend/src/zac/camunda/dynamic_forms/serializers.py
+++ b/backend/src/zac/camunda/dynamic_forms/serializers.py
@@ -2,6 +2,8 @@ from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
 
+from zac.api.polymorphism import PolymorphicSerializer
+
 from ..user_tasks import usertask_context_serializer
 
 FIELD_TYPE_MAP = {
@@ -24,6 +26,7 @@ INPUT_TYPE_MAP = {
 
 class EnumField(serializers.ListField):
     child = serializers.ListField(
+        child=serializers.CharField(),
         label=_("Possible enum choice"),
         help_text=_("First element is the value, second element is the label."),
         min_length=2,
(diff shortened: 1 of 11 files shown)


backend/src/zac/camunda/user_tasks/context.py
```diff
@@ -34,7 +34,14 @@ def get_context(task: Task) -> Optional[Context]:
...
     callable.
     """
-    (callback, *rest) = REGISTRY.get(task.form_key)
+    lookup = task.form_key
+    if task.form_key not in REGISTRY:
+        warnings.warn(
+            f"Unknown task registry key: '{lookup}'. Falling back to dynamic form.",
+            RuntimeWarning,
+        )
+        lookup = ""
```
damm89, line 43: What's your reasoning for not making lookup explicitly "zac:dynamicForm" or whatever? Wouldn't it be clearer for the API docs if it were to be explicit?

backend/src/zac/camunda/dynamic_forms/serializers.py
```diff
@@ -0,0 +1,97 @@
...
+        "int": IntSerializer,
+        "boolean": BooleanSerializer,
+        "date": DatetimeSerializer,
+    }
+
+    name = serializers.CharField(
+        label=_("Field name/identifier"),
+        required=True,
+    )
+    title = serializers.CharField(
```
damm89, line 81: Kelvin and I had agreed on label instead of title, but I guess he can change it back.

damm89 (approved):
Prima!

backend/src/zac/camunda/user_tasks/context.py
sergei-maertens: I'd like to mimic what's the value in Camunda - which is empty normally for these cases.

Figuring out why the value is `zac:dynamicForm` to debug what's happening when you can't find that string anywhere in the process model BPMN would probably be confusing.

backend/src/zac/camunda/dynamic_forms/serializers.py
sergei-maertens: I'll change it, I prefer label too

PR #191, opened 2021-02-09
Added to_internal_value method to polymorphicserializer

damm89:


backend/src/zac/camunda/api/views.py
```diff
@@ -67,16 +67,18 @@ def get(self, request: Request, *args, **kwargs):
         return Response(serializer.data)
 
 
-class GetTaskContextView(APIView):
+class UserTaskView(APIView):  # change into -> UserTaskView
```
damm89, line 70:
```suggestion
class UserTaskView(APIView): 
```

backend/src/zac/camunda/api/views.py
```diff
@@ -67,16 +67,18 @@ def get(self, request: Request, *args, **kwargs):
...
     """
-    Retrieve the user task context from Camunda.
+    Get the user task context from Camunda and perform the user task on Camunda.
 
     Given the task ID, retrieve the task details from Camunda and enrich this with
     context for the UI. The shape of the context depends on the ``form`` value.
+
+    The shape of the payload of the user task depends on the ``form`` value as well.
```
damm89, line 77:
```suggestion
    When submitting user task data, the shape of the payload also depends on the ``form`` value.
```

PR #192, opened 2021-02-09
Integrate PUT on User Task API View and serializers for all relevant user tasks

damm89:


backend/src/zac/camunda/api/urls.py
```diff
@@ -9,7 +14,14 @@
...
-        "task-data/<uuid:task_id>", GetTaskContextView.as_view(), name="get-task-data"
+        "task-data/<uuid:task_id>/submit",
+        SubmitUserTaskView.as_view(),
+        name="submit-task-data",
+    ),
+    path(
+        "task-data/<uuid:task_id>",
+        GetUserTaskContextView.as_view(),
+        name="get-task-data",
     ),
```
sergei-maertens, line 25: I'd stick to using the same URLs:

* `GET` -> get the task data context
* `PUT` -> submit the task data

in the view you can implement `def get_serializer(...)` based on `self.request.method`, and then it should be picked up by the schema generation as well.

backend/src/zac/camunda/api/views.py
```diff
@@ -111,10 +124,10 @@ def get(self, request: Request, task_id: uuid.UUID):
 
     @extend_schema(
         summary=_("Submit user task data"),
-        request=OpenApiTypes.OBJECT,
+        request=SubmitUserTaskSerializer,
         responses={
-            200: OpenApiTypes.OBJECT,
-            400: OpenApiTypes.OBJECT,
+            204: None,
+            400: SubmitUserTaskSerializer,
```
sergei-maertens, line 130: The validation error format is not exactly the same.

It's in my backlog to re-shape validation errors following the shape of https://docs.geostandaarden.nl/api/API-Strategie-ext/#error-handling.

I'd stick to the less helpful but correct `OpenApiTypes.OBJECT`, rather than a schema that doesn't match the actual response.

backend/src/zac/camunda/select_documents/serializers.py
```diff
@@ -0,0 +1,112 @@
...
+
+        return selected_documents
+
+    def get_process_variables(self) -> Dict:
+        """
+        #TODO: ?
+        """
+        return {}
+
+    def on_task_submission(self) -> NoReturn:
```
sergei-maertens, line 108: https://docs.python.org/3/library/typing.html#typing.NoReturn

```suggestion
    def on_task_submission(self) -> None:
```

backend/src/zac/camunda/select_documents/serializers.py
```diff
@@ -0,0 +1,112 @@
...
+    )
+
+    def validate_selected_documents(self, selected_documents):
+        # Make sure selected documents are unique
+        selected_documents = list(dict.fromkeys(selected_documents))
+
+        # Get zaak documents to verify valid document selection
+        task = self.context["task"]
+        process_instance = get_process_instance(task.process_instance_id)
+        self.zaak_url = get_process_zaak_url(process_instance)
```
sergei-maertens, line 85:
```suggestion
        zaak_url = get_process_zaak_url(process_instance)
```
I don't see what this variable should be assigned to `self`?

backend/src/zac/contrib/kownsl/camunda.py
```diff
@@ -54,55 +52,270 @@ class Meta:
...
+        if invalid_docs:
+            raise serializers.ValidationError(
+                _(
+                    "Selected documents: {invalid_docs} are invalid. Please choose one of the "
+                    "following documents: {valid_docs}."
+                ).format(invalid_docs=invalid_docs, valid_docs=valid_docs),
+                code="invalid_choice",
+            )
+
+        return selected_documents
```
sergei-maertens, line 175: This block of code is repeated, would make sense to put that in a `Validator` class/function and specify that validator on the fields in the serializers.

backend/src/zac/contrib/kownsl/camunda.py
```diff
@@ -54,55 +52,270 @@ class Meta:
...
+            ]
+        )
+
+        user_deadlines = {
+            user: str(data["deadline"])
+            for data in self.validated_data["assigned_users"]
+            for user in data["users"]
+        }
+
+        if KownslTypes.advice.lower() in self.context["task"].form_key.lower():
```
sergei-maertens, line 239: this looks incredibly fragile, and it's not obvious what the logic is here.

Is there no better, explicit way?

backend/src/zac/contrib/kownsl/constants.py
```diff
@@ -4,5 +4,5 @@
 
 
 class KownslTypes(DjangoChoices):
-    advice = ChoiceItem("advice", _("Advice"))
-    approval = ChoiceItem("approval", _("Approval"))
+    advice = ChoiceItem("advice", _("Advies"))
+    approval = ChoiceItem("approval", _("Accordering"))
```
sergei-maertens, line 8: This should just have been handled with the translations catalogue :-)

backend/src/zac/contrib/validsign/camunda.py
```diff
@@ -1,23 +1,208 @@
...
+        if invalid_docs:
+            raise serializers.ValidationError(
+                _(
+                    "Selected documents: {invalid_docs} are invalid. Please choose one of the "
+                    "following documents: {valid_docs}."
+                ).format(invalid_docs=invalid_docs, valid_docs=valid_docs),
+                code="invalid_choice",
+            )
+
+        return selected_documents
```
sergei-maertens, line 99: Another candidate for the re-usable validators approach.

backend/src/zac/contrib/validsign/camunda.py
```diff
@@ -1,23 +1,208 @@
...
+                    "firstName": signer_data["first_name"],
+                    "lastName": signer_data["last_name"],
+                }
+            )
+
+        return {
+            "signers": signers,
+        }
+
+    def on_task_submission(self) -> NoReturn:
```
sergei-maertens, line 192:
```suggestion
    def on_task_submission(self) -> None:
```

sergei-maertens (changes requested):
Looks good, couple small remarks

backend/src/zac/contrib/validsign/camunda.py
damm89: Why not a noreturn?

backend/src/zac/contrib/kownsl/camunda.py
damm89: Like a hard-coded mapping...? Not exactly sure. The review_type of the review_request is basically dependent on the form_key.

backend/src/zac/contrib/validsign/camunda.py
sergei-maertens: because it returns `None`, `NoReturn` is meant for callables that don't return at all (because they only raise exceptions for example)

backend/src/zac/api/context.py
```diff
@@ -0,0 +1,33 @@
...
+from zgw_consumers.api_models.zaken import Zaak
+
+from zac.camunda.data import Task
+from zac.camunda.process_instances import get_process_instance
+from zac.core.camunda import get_process_zaak_url
+from zac.core.services import fetch_zaaktype, get_documenten, get_zaak
+
+
+@dataclass
+class ZaakContext:
```
sergei-maertens, line 15: this should probably inherit from `zac.camunda.user_tasks.Context`, otherwise the type hints based on `Context` base class are no longer correct :)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #201, opened 2021-02-15
Integrate dynamic forms

sergei-maertens:
* Implemented dynamic serializer by fetching the serializer fields from the task definition and calling that in the `__new__` method of the base serializer
* Fixed the schema generation to accommodate for arbitrary hash-maps (using `additionalProperties`), which is the case with this dynamic serializer.

I've considered creating a special base class for these HashMap-serializers, however the `drf_spectacular.openapi.resolve_serializer` will break on that because there are no `properties` in the schema, and it would be discarded (see issue https://github.com/tfranzel/drf-spectacular/issues/299).


damm89 (approved):
Cool!

PR #204, opened 2021-02-16
Werkvoorraad endpoints

damm89:
Needs further specification by @kelvincy


backend/src/zac/core/api/permissions.py
```diff
@@ -39,3 +44,12 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
             return False
         return super().has_permission(request, view)
+
+
+class CanHandleAccessRequests(RulesPermission):
+    permission = zaken_handle_access
+
+    def has_permission(self, request: Request, view: APIView) -> bool:
+        if request.method not in permissions.SAFE_METHODS:
+            return False
```
sergei-maertens, line 54: not sure why this is always added? The endpoints only support GET anyway, so hitting this endpoint will return HTTP 405 Method not allowed

backend/src/zac/werkvoorraad/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+        help_text=_("Username of access requester/grantee"),
+    )
+
+    class Meta:
+        model = AccessRequest
+        fields = ("requester",)
+
+
+class WorkStackAccessRequestsSerializer(APIModelSerializer):
+    requesters = AccessRequesterSerializer(many=True)
```
sergei-maertens, line 33:
```suggestion
    access_requests = AccessRequestSerializer(many=True)
```

nit-picky, but "access requester" makes me think about just the user behind the access request, not the access request itself. Also keep in mind if we add attributes to the access request and have to expose them in the API in the future, it would look semantically incorrect to have this keyed under `requesters`.

backend/src/zac/werkvoorraad/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+    url = serializers.SerializerMethodField()
```
sergei-maertens, line 34: would benefit from a help text. I'd also mark it deprecated in the schema (I think that's possible via drf-spectacular), because I expect this to be removed once the UI has the interface to manage access requests.

backend/src/zac/werkvoorraad/data.py
```diff
@@ -0,0 +1,20 @@
...
+@dataclass
+class ActivityGroup:
+    activities: List[Activity]
+    zaak: Zaak
+    zaak_url: str
+
+
+@dataclass
+class AccessRequestGroup:
+    requesters: List[AccessRequest]
```
sergei-maertens, line 18:
```suggestion
    requests: List[AccessRequest]
```
or `access_requests` is a better name, see the serializer remarks

backend/src/zac/werkvoorraad/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+
+class ActivityNameSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Activity
+        fields = ("name",)
+
+
+class WorkStackAdhocActivitiesSerializer(APIModelSerializer):
+    activities = ActivityNameSerializer(many=True)
+    url = serializers.SerializerMethodField()
```
sergei-maertens, line 63: Help text + deprecation would be useful

backend/src/zac/werkvoorraad/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+                "bronorganisatie": obj.zaak.bronorganisatie,
+                "identificatie": obj.zaak.identificatie,
+            },
+        )
+
+
+class CaseTypeDescriptionSerializer(APIModelSerializer):
+    class Meta:
+        model = ZaakType
+        fields = ("omschrijving",)
```
sergei-maertens, line 87: I'd rather re-use `zac.core.api.serializers.ZaakTypeSerializer`

backend/src/zac/werkvoorraad/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+        )
+
+    def get_url(self, obj) -> str:
+        return reverse(
+            "core:zaak-detail",
+            kwargs={
+                "bronorganisatie": obj.bronorganisatie,
+                "identificatie": obj.identificatie,
+            },
+        )
```
sergei-maertens, line 113: I'd also consider using `zac.core.api.serializers.ZaakDetailSerializer` here rather than defining a new serializer. The `url` is not useful here - Kelvin can build the frontend URL using bronorganisatie + identificatie already.

backend/src/zac/werkvoorraad/serializers.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+class WorkStackUserTaskSerializer(APIModelSerializer):
```
sergei-maertens, line 116: I'd re-use `zac.camunda.api.serializers.TaskSerializer` here and drop `url` - Kelvin can route to the relevant page I think. Possibly we need to serialize some zaak-information as well to be able to link to the zaak detail page.

backend/src/zac/werkvoorraad/views.py
```diff
@@ -2,19 +2,33 @@
 from typing import List
 
 from django.contrib.auth.mixins import LoginRequiredMixin
+from django.utils.translation import gettext as _
 from django.views.generic import TemplateView
 
 from django_camunda.camunda_models import Task, factory
 from django_camunda.client import get_client
+from rest_framework import authentication, permissions, views as drf_views
```
sergei-maertens, line 10: Please create a subpackage `api` for the API views, serializers etc... in `werkvoorraad`. Let's keep API views & classic Django views separated.

backend/src/zac/werkvoorraad/views.py
```diff
@@ -58,6 +72,81 @@ def get_access_requests_groups(user: User):
...
+    def get_serializer(self, **kwargs):
+        return WorkStackAccessRequestsSerializer(many=True, **kwargs)
+
+    def get(self, request: Request) -> Response:
+        access_requests_groups = get_access_requests_groups(request.user)
+        access_requests_groups = [
+            AccessRequestGroup(**group) for group in access_requests_groups
+        ]
+        serializer = self.get_serializer(instance=access_requests_groups)
+        return Response(serializer.data)
```
sergei-maertens, line 89: I think you can let this inherit from `generics.ListView` and override `def get_queryset` to return the `access_requests_groups`. That should help a great deal with reducing the boilerplate & allows you to just define `serializer_class` and be done.

Alternatively, we could set up our own `List` base class.

backend/src/zac/werkvoorraad/views.py
```diff
@@ -58,6 +72,81 @@ def get_access_requests_groups(user: User):
     return requested_zaken
 
 
+class WorkStackAccessRequestsView(drf_views.APIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAuthenticated & CanHandleAccessRequests,)
+    schema_summary = _("List access requests")
```
sergei-maertens, line 78: I forgot to mention this, but that's a bit of code that will be removed. It's better to use `@drf_spectacular.utils.extend_schema(summary=_("..."))`

sergei-maertens (changes requested):
Some polishing required :)

backend/src/zac/werkvoorraad/serializers.py
damm89: But isn't that against ZGW? Request only data you need.

backend/src/zac/core/api/permissions.py
damm89: Also not sure :-)

backend/src/zac/werkvoorraad/tests/test_user_tasks_endpoint.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def _get_task(**overrides):
+    data = underscoreize({**TASK_DATA, **overrides})
+    return factory(Task, data)
+
+
+class UserTasksTests(APITestCase):
+    """
+    Test the adhoc activities API endpoint.
```
damm89, line 45:
```suggestion
    Test the user tasks workstack API endpoint.
```

backend/src/zac/werkvoorraad/serializers.py
sergei-maertens: we're getting the data from the upstream endpoint anyway, it's part of the resource. I wouldn't say it's a violation?

sergei-maertens (approved):

sergei-maertens:
Marking as ready - if CI passes, we'll merge. If Kelvin has feedback, that can be done in a separate PR :)

PR #207, opened 2021-02-18
Fix/submit user task

damm89:


backend/src/zac/camunda/api/views.py
```diff
@@ -76,7 +77,6 @@ class UserTaskView(APIView):
     """
 
     permission_classes = (permissions.IsAuthenticated & CanPerformTasks,)
-    parser_classes = (parsers.JSONParser,)
```
sergei-maertens, line 79: This breaks the camelCase dynamic form variable names, because they are converted to snake_case now. We can't transform this at all, since there may also be snake_case variables in the processes, so always converting to camelCase would break that.

backend/src/zac/camunda/api/views.py
```diff
@@ -150,6 +149,13 @@ def put(self, request: Request, task_id: uuid.UUID):
             **serializer.get_process_variables(),
         }
 
+        # Make sure variables are rendered in preferred format
+        # before being passed on to complete_task.
+        if not getattr(request, "accepted_renderer", None):
+            neg = self.perform_content_negotiation(request, force=True)
+            request.accepted_renderer, request.accepted_media_type = neg
+
+        variables = json.loads(request.accepted_renderer.render(variables))
```
sergei-maertens, line 158: can't do that, our renderer has no connection with the format of the expected process variables

backend/src/zac/camunda/api/views.py
damm89: What do you suggest? Currently the frontends needs to submit the variables snake_case.

sergei-maertens: not sure yet, I'll have to take a look. It's annoying, because the schema generation also applies the post-processing globally rather than taking each view(set) parser into account

backend/src/zac/camunda/user_tasks/context.py
```diff
@@ -1,16 +1,30 @@
...
+Renderers = Tuple[Type[renderers.BaseRenderer]]
+
+
+@dataclass
+class RegistryItem:
+    callback: callable
+    read_serializer: SerializerCls
+    write_serializer: Optional[SerializerCls] = None
+    parsers: Optional[Parsers] = None
+    renderers: Optional[Renderers] = None
```
damm89, line 24: Yeah this is nicer.

PR #209, opened 2021-02-18
access to zaak for connected users

### LINKED ISSUE #183: Authorizations: allow users who are connected with zaak to see it
See details at https://github.com/GemeenteUtrecht/ZGW/issues/682

Change permission (add a rule predicate), that if the user has one of the following roles in the zaak, they can always see it:
* behandelaar, 
* uitgenodigde medewerker
* adviseurs/accordeurs

annashamray:
fixes #183

### DIFF
--- a/backend/src/zac/contrib/kownsl/tests/test_zaak_review_requests.py
+++ b/backend/src/zac/contrib/kownsl/tests/test_zaak_review_requests.py
@@ -460,16 +460,6 @@ class ZaakReviewRequestsPermissionTests(ClearCachesMixin, APITestCase):
         self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
     def test_rr_detail_authenticated_no_permissions(self):
-        # mock_service_oas_get(m, ZAKEN_ROOT, "zrc")
-        # mock_service_oas_get(m, KOWNSL_ROOT, "kownsl", oas_url=f"{KOWNSL_ROOT}api/v1")
-        # m.get(
-        #     f"{ZAKEN_ROOT}rollen?zaak={self.zaak['url']}",
-        #     json=paginated_response([]),
-        # )
-        # m.get(
-        #     f"{KOWNSL_ROOT}api/v1/review-requests?for_zaak={self.zaak['url']}",
-        #     json=[],
-        # )
         user = UserFactory.create()
         self.client.force_authenticate(user=user)
 
--- a/backend/src/zac/contrib/kownsl/tests/test_zaak_review_requests.py
+++ b/backend/src/zac/contrib/kownsl/tests/test_zaak_review_requests.py
@@ -31,6 +31,7 @@ from zgw.models.zrc import Zaak
 CATALOGI_ROOT = "http://catalogus.nl/api/v1/"
 ZAKEN_ROOT = "http://zaken.nl/api/v1/"
 DOCUMENTS_ROOT = "http://documents.nl/api/v1/"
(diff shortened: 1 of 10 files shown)


backend/src/zac/core/rules.py
```diff
@@ -189,6 +189,19 @@ def is_zaak_behandelaar(user: User, zaak: Optional[Zaak]):
     return bool(user_rollen)
 
 
+@rules.predicate
+def is_zaak_adviser(user: User, zaak: Optional[Zaak]):
```
sergei-maertens, line 193:
```suggestion
def is_zaak_reviewer(user: User, zaak: Optional[Zaak]):
```
reviewer is the more generic form of advisor/approver

sergei-maertens (changes requested):
Stylistic changes, logically this looks solid.

backend/src/zac/core/rules.py
```diff
@@ -189,6 +189,19 @@ def is_zaak_behandelaar(user: User, zaak: Optional[Zaak]):
...
+    from zac.contrib.kownsl.api import get_review_requests
+
+    if zaak is None:
+        return False
```
damm89, line 197: Do you really want False? Shouldn't the predicate just be ignored? (i.e. shouldn't it rather return None)

backend/src/zac/core/rules.py
```diff
@@ -178,7 +178,7 @@ def is_zaak_behandelaar(user: User, zaak: Optional[Zaak]):
     from .services import get_rollen
 
     if zaak is None:
-        return True
+        return False
```
damm89, line 181: Same as below: Do you really want False? Shouldn't the predicate just be ignored? (i.e. shouldn't it rather return None)

damm89 (commented):
Looks good - just a question.

backend/src/zac/contrib/kownsl/tests/test_zaak_review_requests.py
```diff
@@ -440,21 +441,53 @@ def test_rr_detail_not_authenticated(self):
...
+        # mock_service_oas_get(m, ZAKEN_ROOT, "zrc")
+        # mock_service_oas_get(m, KOWNSL_ROOT, "kownsl", oas_url=f"{KOWNSL_ROOT}api/v1")
+        # m.get(
+        #     f"{ZAKEN_ROOT}rollen?zaak={self.zaak['url']}",
+        #     json=paginated_response([]),
+        # )
+        # m.get(
+        #     f"{KOWNSL_ROOT}api/v1/review-requests?for_zaak={self.zaak['url']}",
+        #     json=[],
+        # )
```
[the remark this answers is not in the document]
annashamray: or, they are remnants of my desperate attempts to mate all tests work
I'll remove it

backend/src/zac/core/rules.py
annashamray: I added this condition to make it consistent with other rule predicates. I think that this condition should be removed since these rules should not be applied when `zaak is None`. Perhaps this could be handled during auth refactoring,

backend/src/zac/core/rules.py
annashamray: Please, take a look at my comment above

backend/src/zac/core/rules.py
sergei-maertens: Thanks, makes sense, as this predicate really has no meaning without `zaak` object.

sergei-maertens (approved):

PR #223, opened 2021-03-04
Added review documents to review request details on zaak detail page …

damm89:
…and fixed the polymorphic revreqdetail serializer.

Pertains to https://github.com/GemeenteUtrecht/ZGW/issues/702 - first checkbox.

### DIFF
--- a/backend/src/zac/contrib/dowc/views.py
+++ b/backend/src/zac/contrib/dowc/views.py
@@ -2,7 +2,8 @@ from typing import Any, Optional
 
 from django.utils.translation import gettext_lazy as _
 
-from drf_spectacular.utils import extend_schema
+from drf_spectacular.types import OpenApiTypes
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework import authentication, permissions
 from rest_framework.response import Response
 from rest_framework.views import APIView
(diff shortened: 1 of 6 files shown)


backend/src/zac/contrib/dowc/views.py
```diff
@@ -26,25 +26,26 @@ def _cast(value: Optional[Any], type_: type) -> Any:
 
 @extend_schema(summary=_("Open document for viewing or editing"))
 class OpenDowcView(APIView):
+    """
+    You can pass the "versie" as a query parameter to specify the document version.
+    """
+
```
annashamray, line 32: I think it's nice to see `versie` rendered as query-param in the OAS. There are already 3 ways used to do it in ZAC:
* use @extend_schema(parameters=[OpenApiParameter(name="versie", ... )])` like in `InformatieObjectTypeListView`
* use input serializer for query params like in `GetZakenView`
* use `ApiFilterSet` like in `ZaakTypenView`

Perhaps we need to make guidelines on how to generate OAS for endpoints among backend developers and stick to them because now it's chaotic and reduces readability.

damm89: Yes, sounds good! Thanks for your review.

damm89: Btw: I like the @extend_schema one the best. I think it's the cleanest and most explicit implementation.

annashamray (approved):

PR #231, opened 2021-03-10
Feature/auth permission rules

### LINKED ISSUE #225: Auth refactorig: create PermissionDefinition model and set up model layer
Parent issue - #224

* [x] (#225) create base `Policy` (or `Blueprint`) class which will include serializer to validate `policy` data and a function to check permissions for a particular object (zaak/doc). Add it to `Permission` class
    The structure should be like this:
  ```yaml
  Permission:
      name
      description
      policy:
          blueprint_serializer
          has_access
  ```

* [x] create model `PermissionDefinition`. See detail in https://github.com/GemeenteUtrecht/ZGW/pull/685. When the data is saved `policy` field should be validated on `blueprint_serializer`. 
* [x] `User` model should have m2m relations with `PermissionDefinition`. `User` model should have `permission_definitions` method/property which gathers all related permission definitions including those in `AuthorizationProfile` objects.
* [x] In the `PermissionDefinition` admin page a description of the `policy` field should be displayed, which depends on the value of the `permission` attribute. (Nice to have)
* ~~remove old code~~ Upd. old code would be removed after new permissions for all apps are created

annashamray:
fixes #225


annashamray:
wrong target branch

Issue #224, opened 2021-03-04
Authorization: refactoring


annashamray:
Parent user story for all refactoring tasks

Concept: https://github.com/GemeenteUtrecht/ZGW/pull/685

**Approaches:**
We have two options:
1. use one model to store both blueprint and atomic permissions
2. use two separate models to store them.

In order not to do the same work twice, we can try both approaches sequentially: first, we can build a working POC with the first approach and then tweak it to explore the second option.  

**Technical steps:**
* [x] Model layer (#225) 
    - [x] create base `Policy` (or `Blueprint`) class which will include serializer to validate `policy` data and a function to check permissions for a particular object (zaak/doc). Add it to `Permission` class
    The structure should be like this:
  ```yaml
  Permission:
      name
      description
      policy:
          blueprint_serializer
          has_access
  ```

    - [x] create model `PermissionDefinition`. See detail in https://github.com/GemeenteUtrecht/ZGW/pull/685. When the data is saved `policy` field should be validated on `blueprint_serializer`. 
    - [x]  `User` model should have m2m relations with `PermissionDefinition`. `User` model should have `permission_definitions` method/property which gathers all related permission definitions including those in `AuthorizationProfile` objects.
    - [x] In the `PermissionDefinition` admin page a description of the `policy` field should be displayed, which depends on the value of the `permission` attribute. (Nice to have)

* [x] (#226) Rule predicates
There are two options how to refactor rule predicates:
1. Permissions are defined only on `PermissionDefinition` data and rules are not needed. For example, if the user is an advisor and they need access to zaak-detail page, the atomic `PermissionDefinition` object would be created with permissions for a particular zaak. It can be done automatically using notifications or manually by administrators. In this case, the rules backend can be removed. Permission check can be performed only in permission mixin. 
2. We will have permissions defined both in `PermissionDefinition` and rules. In this case, a generic rule should check if the user has atomic access or policy access to the object. All rules should be tested only when the object (zaak or doc) exists.

@sergei-maertens Which options do you prefer? I like the first one better because in this case, we have all permission definitions in one place which can be conveniently accessed via DB queries 

Upd. the 1st option has been chosen

* [x] (#227) Filtering list of zaken.
    - [x] add a method `filter_es (search: Search) -> Search` to `Policy` class to update an ES search on the policy blueprint. 
    - [x] In the ES search iterate over all `PermissionDefinition` objects with blueprints and use their `filter_es` methods
[...]

PR #232, opened 2021-03-10
Feature/auth permission rules

### LINKED ISSUE #226: Auth refactoring: refactor permissions used in views
Parent issue - #224

 Permissions are defined only on `PermissionDefinition` data and rules are not needed. For example, if the user is an advisor and they need access to zaak-detail page, the atomic `PermissionDefinition` object would be created with permissions for a particular zaak. It can be done automatically using notifications or manually by administrators. In this case, the rules backend can be removed. Permission check can be performed only in permission mixin. 
* [x] create `PermissionMixin` which doesn't use rules backend and checks permissions based on `PermissionDefinition`
* [x] add subclasses of `Blueprint` for zaken and docs
* [x] add blueprint classes to permissions
* [x] migrate data from old auth models to the new ones
* [x] create management command to create atomic permission definitions instead of dynamic rules
* [x] add notification handlers to update atomic permissions

annashamray:
fixes #226

depends on #228, https://github.com/GemeenteUtrecht/kownsl/pull/32

### DIFF
--- a/backend/src/zac/api/permissions.py
+++ b/backend/src/zac/api/permissions.py
@@ -85,17 +85,18 @@ class DefinitionBasePermission(permissions.BasePermission):
         permission_definitions = (
             PermissionDefinition.objects.for_user(request.user)
             .actual()
-            .filter(permission=self.permission)
+            .filter(permission=self.permission.name)
         )
 
-        # first check atomic permissions
+        # first check atomic permissions - this checks both atomic permissions directly attached to the user
+        # and atomic permissions defined to authorization profiles
         if permission_definitions.filter(object_url=obj.url).exists():
             return True
 
         # then check blueprint permissions
         for permission in (
             permission_definitions.filter(object_type=self.object_type)
-            .exclude(policy={})
+            .filter(object_url="")
             .all()
         ):
             if permission.has_policy_access(obj, request):
--- a/backend/src/zac/api/permissions.py
+++ b/backend/src/zac/api/permissions.py
@@ -84,8 +84,8 @@ class DefinitionBasePermission(permissions.BasePermission):
     def has_object_permission(self, request, view, obj):
         permission_definitions = (
             PermissionDefinition.objects.for_user(request.user)
+            .filter(permission=self.permission.name, object_type=self.object_type)
             .actual()
-            .filter(permission=self.permission.name)
         )
 
         # first check atomic permissions - this checks both atomic permissions directly attached to the user
@@ -94,11 +94,7 @@ class DefinitionBasePermission(permissions.BasePermission):
             return True
 
         # then check blueprint permissions
-        for permission in (
-            permission_definitions.filter(object_type=self.object_type)
-            .filter(object_url="")
-            .all()
-        ):
+        for permission in permission_definitions.filter(object_url=""):
             if permission.has_policy_access(obj, request):
                 return True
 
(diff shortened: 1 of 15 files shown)


backend/src/zac/api/permissions.py
```diff
@@ -65,3 +67,38 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+            raise ImproperlyConfigured(
+                "%s is missing the 'permission' attribute" % cls.__name__
+            )
+        return super().__new__(cls, *args, **kwargs)
+
+    def has_object_permission(self, request, view, obj):
+        permission_definitions = (
+            PermissionDefinition.objects.for_user(request.user)
+            .actual()
+            .filter(permission=self.permission)
```
sergei-maertens, line 88:
```suggestion
            .filter(permission=self.permission.name)
```
I think

backend/src/zac/api/permissions.py
```diff
@@ -65,3 +67,38 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+        )
+
+        # first check atomic permissions
+        if permission_definitions.filter(object_url=obj.url).exists():
+            return True
```
sergei-maertens, line 93:
```suggestion
        # first check atomic permissions - this checks both atomic permissions directly attached to the user
        # and atomic permissions defined to authorization profiles
        if permission_definitions.filter(object_url=obj.url).exists():
            return True
```

backend/src/zac/api/permissions.py
```diff
@@ -65,3 +67,38 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+
+        # then check blueprint permissions
+        for permission in (
+            permission_definitions.filter(object_type=self.object_type)
+            .exclude(policy={})
```
sergei-maertens, line 98: Since we're not dealing with atomic permissions here anymore, and `policy` can be "empty" but also `null` (see your models and review of other PR), this looks a bit odd. How are values handled that are `null`? And - if there's no `object_url`, that makes it a policy definition because of the mutual exclusivity, so shouldn't that mean that a policy MUST be entered?

```suggestion
            .filter(object_url="")
```

backend/src/zac/api/permissions.py
```diff
@@ -65,3 +67,38 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
+            .all()
```
sergei-maertens, line 99: not needed if you already have a `.filter()` or `.exclude` call

backend/src/zac/notifications/handlers.py
```diff
@@ -34,6 +35,9 @@ def handle(self, data: dict) -> None:
 
         elif data["resource"] == "rol":
             self._handle_rol_change(data["hoofd_object"])
+            # TODO should we remove permission if the rol is deleted?
```
sergei-maertens, line 38: Can we bundle these questions somewhere to ask them to the rest of the team? Or maybe ask the question on Slack in the general channel.

sergei-maertens (changes requested):
This is coming together nicely!

backend/src/zac/api/permissions.py
annashamray: you are right, thanks for explanation

backend/src/zac/api/permissions.py
annashamray: `.all()` is removed, `.exclude` is replaced with `.filter(object_url="")`

backend/src/zac/notifications/handlers.py
annashamray: I'll add this question along with future questions to #224 and I'll ask them on Friday meeting, I think there should be some introduction to the context of this question :)

sergei-maertens (approved):

PR #233, opened 2021-03-11
Added case result to zaakdetail api view

damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/702

### DIFF
--- a/backend/src/zac/core/api/serializers.py
+++ b/backend/src/zac/core/api/serializers.py
@@ -198,6 +198,7 @@ class ZaakDetailSerializer(APIModelSerializer):
             "vertrouwelijkheidaanduiding",
             "deadline",
             "deadline_progress",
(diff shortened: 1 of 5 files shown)


backend/src/zac/core/api/serializers.py
```diff
@@ -198,6 +198,7 @@ class Meta:
             "vertrouwelijkheidaanduiding",
             "deadline",
             "deadline_progress",
+            "resultaat",
```
sergei-maertens, line 201: This will only include the `resultaat` URL reference, no? You'll need a nested serializer for the `resultaat`, as evidenced by the missing test where it's not `None` as well :-)

sergei-maertens (changes requested):

sergei-maertens (approved):

PR #245, opened 2021-03-17
Feature/auth endpoints meta

annashamray:
fixes #239
depends on #244

PR #249, opened 2021-03-18
Feature/auth endpoints core

### LINKED ISSUE #240: Auth refactoring: core app
parent issue - #224

* [x] update all bff endpoints with a new permission mixin
* [x] update all the tests

annashamray:
fixes #240
depends on #245

### DIFF
--- a/backend/src/zac/api/permissions.py
+++ b/backend/src/zac/api/permissions.py
@@ -82,6 +82,9 @@ class DefinitionBasePermission(permissions.BasePermission):
         return super().__new__(cls, *args, **kwargs)
 
     def has_object_permission(self, request, view, obj):
+        if request.user.is_superuser:
+            return True
+
         permission_definitions = (
             PermissionDefinition.objects.for_user(request.user)
             .filter(permission=self.permission.name, object_type=self.object_type)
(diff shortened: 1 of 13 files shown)


backend/src/zac/api/permissions.py
```diff
@@ -82,6 +82,9 @@ def __new__(cls, *args, **kwargs):
         return super().__new__(cls, *args, **kwargs)
 
     def has_object_permission(self, request, view, obj):
```
damm89, line 84: I know this was done in an earlier PR, so it kinda sucks I didn't pick up on it then but line 100 seems a bit opaque in what you're actually querying. Wouldn't it be nicer if the permission_definitions queryset had a method attached that would get all blueprint permissions (i.e., do the filter behind the scenes)?

backend/src/zac/api/permissions.py
```diff
@@ -123,15 +129,15 @@ def has_permission(self, request: Request, view: APIView) -> bool:
...
-
         serializer = view.get_serializer(data=request.data)
         # if the serializer is not valid, we want to see validation errors -> permission is granted
         if not serializer.is_valid():
             return True
 
+        # first check  if the user has permissions for any object
+        if super().has_permission(request, view) is False:
+            return False
+
```
damm89, line 140: I think PEP 8 states this should probably be rewritten as:

```suggestion
        if not super().has_permission(request, view):
            return False

```

damm89 (changes requested):

damm89:
Very nice btw. It's like wearing new shoes when I see all the new and improved DefinitionBasePermissions & ZaakDefinitionPermissions flying around.

SilviaAmAm (approved):

sergei-maertens (approved):

annashamray:
@damm89 Thank you very much, I'm happy that now everything is getting more sense

PR #257, opened 2021-03-19
Fixed a 'major' lack of functionality in user task submission - hope …

damm89:
…this works now.

### DIFF
--- a/backend/src/zac/camunda/api/serializers.py
+++ b/backend/src/zac/camunda/api/serializers.py
@@ -65,7 +65,7 @@ class BaseUserTaskSerializer(PolymorphicSerializer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields["form"].choices = [
-            (key, key or _("(camunda form)")) for key in self.serializer_mapping.keys()
+            (key, key or _("(camunda form)")) for key in REGISTRY.keys()
         ]
 
 
--- a/backend/src/zac/camunda/api/serializers.py
+++ b/backend/src/zac/camunda/api/serializers.py
@@ -1,5 +1,4 @@
-from collections import defaultdict
-from typing import Any, Dict, List, NoReturn
+from typing import Any, Dict, List
 
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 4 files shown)


backend/src/zac/camunda/api/serializers.py
```diff
@@ -39,12 +40,18 @@ class ProcessInstanceSerializer(serializers.Serializer):
     tasks = TaskSerializer(many=True)
 
 
+class ChoiceFieldNoValidation(serializers.ChoiceField):
+    def to_internal_value(self, data):
+        if not isinstance(data, str):
+            raise serializers.ValidationError("Form needs to be a string.")
+
```
sergei-maertens, line 47: Shouldn't this return `data`?

backend/src/zac/camunda/api/serializers.py
```diff
@@ -76,28 +83,37 @@ def __init__(self, *args, **kwargs):
...
+        lookup = (
+            form_key
+            if form_key in self.serializer_mapping
+            else self.fallback_distriminator_value
+        )
+        return self.serializer_mapping[lookup]
+
+    def is_valid(self, raise_exception=True):
+        super().is_valid(raise_exception=raise_exception)
+        self.get_mapped_serializer().is_valid(raise_exception=raise_exception)
```
sergei-maertens, line 103: I'd expect this to be the job of the `PolymorphicSerializer` base class, no?

sergei-maertens (changes requested):
Looks okay, but I'm not sure if this is a problem in the base class or this specific user task serializer?

I'd also like to see some regression tests that fail without this patch and pass with the patch.

damm89:
> Looks okay, but I'm not sure if this is a problem in the base class or this specific user task serializer?
> 
> I'd also like to see some regression tests that fail without this patch and pass with the patch.

Basically the only problem was that the ChoiceField for the form key was throwing validation errors...

sergei-maertens (approved):

sergei-maertens:
I'd still expect a regression test though!

sergei-maertens (approved):

PR #261, opened 2021-03-22
DocumentSelectTaskSerializer now includes document type

damm89:
Dependent on https://github.com/GemeenteUtrecht/ZGW/pull/738

### DIFF
--- a/backend/src/zac/core/api/views.py
+++ b/backend/src/zac/core/api/views.py
@@ -25,12 +25,9 @@ from rest_framework.generics import ListAPIView
 from rest_framework.request import Request
 from rest_framework.response import Response
 from zgw_consumers.api_models.base import factory
-from zgw_consumers.api_models.catalogi import InformatieObjectType
 from zgw_consumers.api_models.zaken import Zaak
 from zgw_consumers.concurrent import parallel
-from zgw_consumers.constants import APITypes
 from zgw_consumers.models import Service
-from zgw_consumers.service import get_paginated_results
 
 from zac.accounts.permissions import UserPermissions
 from zac.contrib.brp.api import fetch_extrainfo_np
@@ -449,7 +446,7 @@ class CreateZaakDocumentView(views.APIView):
     parameters=[
         OpenApiParameter(
             name="zaak",
-            required=False,
+            required=True,
             type=OpenApiTypes.URI,
             description=_("Zaak to list available document types for"),
             location=OpenApiParameter.QUERY,
@@ -458,9 +455,7 @@ class CreateZaakDocumentView(views.APIView):
 )
 class InformatieObjectTypeListView(generics.ListAPIView):
     """
-    List the available document types.
-
-    A zaak url can be provided to filter on available document types for a specific zaak.
+    List the available document types for a given zaak.
     TODO: permissions checks on zaak - can this user read/mutate the zaak?
     """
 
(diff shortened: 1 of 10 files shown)


backend/src/zac/core/camunda/select_documents/serializers.py
```diff
@@ -53,14 +53,52 @@ class Meta:
 #
 
 
+class SelectedDocumentSerializer(serializers.Serializer):
+    document = serializers.ChoiceField(
+        child=serializers.URLField,
```
damm89, line 58:
```suggestion
```

backend/src/zac/core/camunda/select_documents/serializers.py
```diff
@@ -53,14 +53,51 @@ class Meta:
...
 
+class SelectedDocumentSerializer(serializers.Serializer):
+    document = serializers.ChoiceField(
+        label=_("Selected document"),
+        help_text=_("The URL of the selected document from the relevant case."),
+        choices=(),
+    )
+
+    document_type = serializers.URLField(
+        child=serializers.URLField,
```
damm89, line 64:
```suggestion
    document_type = serializers.ChoiceField(
```

backend/src/zac/core/api/views.py
```diff
@@ -456,7 +456,6 @@ def post(self, request: Request) -> Response:
 class InformatieObjectTypeListView(generics.ListAPIView):
     """
     List the available document types for a given zaak.
-
```
sergei-maertens, line 459: PEP 257 asks you to format docstrings like this:

```
A short summary, preferably < 70-ish characters.

A more elaborate description, following a blank line.
```

This has an impact on how editors, documentation-generation tools etc. show class/function documentation.

backend/src/zac/core/camunda/select_documents/serializers.py
```diff
@@ -53,27 +55,107 @@ class Meta:
...
+        self.zaak = self.get_zaak_from_context()
+
+        # Validate selected documents
+        doc_urls = [doc["document"] for doc in selected_docs]
+        validate_zaak_documents(doc_urls, self.zaak)
+
+        # Validated selected document types according to case type
+        ztc_client = _client_from_url(self.zaak.zaaktype)
+        results = get_paginated_results(ztc_client, "informatieobjecttype")
+        valid_eiots = [iot["url"] for iot in results]
```
sergei-maertens, line 92: This is not the set restricted to the zaaktype - for that you need to check `zaaktype.informatieobjecttypen` (see https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/VNG-Realisatie/catalogi-api/1.0.0/src/openapi.yaml#operation/zaaktype_read)

sergei-maertens (changes requested):

backend/src/zac/core/api/views.py
damm89: Accidentally edited this!

backend/src/zac/core/camunda/select_documents/serializers.py
damm89: Good catch

sergei-maertens (approved):
I'm expecting some rework once Kelvin gets started with this, since it's not clear for him now which informatieobjecttypen should be shown in the UI and which don't.

https://zac-test.utrechtproeftuin.nl/api/docs/#operation/core_document_types_list is the closest we have, but not sufficient, because we should support `zaaktype` querystring here as well, and pass `zaaktype` to the frontend so it can be used to query.

PR #263, opened 2021-03-23
Added case details to user tasks in workstack

damm89:


backend/src/zac/werkvoorraad/api/serializers.py
```diff
@@ -77,3 +79,25 @@ def get_url(self, obj) -> str:
...
+        fields = (
+            "bronorganisatie",
+            "identificatie",
+            "url",
+        )
+
+
+class WorkStackTaskSerializer(APIModelSerializer):
+    task = TaskSerializer()
+    zaak = CaseSerializer()
```
annashamray, line 96: I think you can reuse `ZaakSerializer` which is already imported in this file

damm89: Yeah, but I didn't like all the unnecessary fields. I'll switch it then.

annashamray (approved):

sergei-maertens (approved):

PR #273, opened 2021-03-29
Auth refactor: approach 1

sergei-maertens:


annashamray:
the second approach for refactoring has been chosen

PR #279, opened 2021-04-01
Feature/admin UI for auth

### LINKED ISSUE #270: Implement interface to manage authorization profiles
With the permissions refactor to use blueprints/atomic permissions, we no longer have an interface to manage the rights of authorization profiles (deleted with #255, but it was no longer functional anyway after the refactor).

We need to design API endpoints and a frontend interface to manage these.

annashamray:
fixes #270
depends on https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/281


sergei-maertens:
@annashamray you can rebase on main :-) #281 is merged

annashamray:
@sergei-maertens thanks for merging authorizations! The rebase is made

backend/src/drf_jsonschema/convert.py
```diff
@@ -0,0 +1,46 @@
...
+            continue
+        sub_schema = field_to_jsonschema(field)
+        if field.required:
+            required.append(name)
+        properties[name] = sub_schema
+
+    result = {"type": "object", "properties": properties}
+    if required:
+        result["required"] = required
+    return result
```
damm89, line 46: I'm assuming type hints aren't as useful here because of the many different shapes?

backend/src/zac/accounts/models.py
```diff
@@ -251,7 +254,12 @@ class Meta:
         verbose_name_plural = _("blueprint definitions")
 
     def __str__(self):
-        return f"{self.permission} ({self.object_type})"
+        if not self.permission or not self.policy:
+            return f"{self.permission}"
+
+        blueprint_class = self.get_blueprint_class()
+        blueprint = blueprint_class(self.policy)
+        return f"{self.permission}: {blueprint.short_display()}"
```
damm89, line 262: This seems funky. Could you explain what's returned here if 'not self.permission is True'?

damm89 (commented):
:+1:

backend/src/drf_jsonschema/convert.py
annashamray: I've copied with code from https://github.com/isprojects/drf_jsonschema without changing anything. I couldn't use it as an external lib, since it's not deployed at PyPI
I'll add type hints to make the code clear

backend/src/zac/accounts/models.py
annashamray: this `if` is for creating new objects. When `self.permission` is empty we can't extract `blueprint_class` and the code can crash. So when the object is not created yet `__str__` will return en empty string, but it's ok

backend/src/drf_jsonschema/converters.py
```diff
@@ -0,0 +1,243 @@
...
+    format = "date-time"
+
+    expected_input_formats = None
+    settings_format = None
+    settings_input_formats = None
+
+    def convert(self, field):
+        # ugh had to copy from DRF
+        format = getattr(self, "format", self.settings_format)
+        # FIXME: I mysteriously get out 'date-time' instead of 'iso-8601'...
```
damm89, line 130: Is this still applicable? I don't think so right?

annashamray: haha, I didn't notice it when copypasting this code. I'll remove it

backend/src/drf_jsonschema/converters.py
```diff
@@ -0,0 +1,243 @@
...
+            type = self.type
+        return {"type": type}
+
+
+class FormatConverter(Converter):
+    type = "string"
+    format = None
+
+    def convert(self, field):
+        result = super(FormatConverter, self).convert(field)
```
damm89, line 29: In python 3 we can just do: 

super().some_method() 

or are we also supporting python 2?

damm89: Don't have to change it for me, btw. I just noticed it.

damm89 (approved):
Good for me!

backend/src/drf_jsonschema/converters.py
annashamray: I wish this lib was on PyPI :) Perhaps the creators of https://github.com/isprojects/drf_jsonschema wanted to support python 2 and I just copied it

PR #282, opened 2021-04-08
FE now gets the available document types from the related case type o…

damm89:
…f the related case

### DIFF
--- a/backend/src/zac/core/camunda/select_documents/context.py
+++ b/backend/src/zac/core/camunda/select_documents/context.py
@@ -19,9 +19,9 @@ from .serializers import (
 def get_context(task: Task) -> DocumentSelectContext:
     zaak_context = get_zaak_context(task, require_documents=True)
     process_instance = get_process_instance(task.process_instance_id)
-    bijdrage_zaaktype_url = process_instance.get_variable("zaaktype")
-    bijdrage_zaaktype = fetch_zaaktype(bijdrage_zaaktype_url)
-    informatieobjecttypen = get_informatieobjecttypen(bijdrage_zaaktype.catalogus)
+    related_zaaktype_url = process_instance.get_variable("zaaktype")
+    related_zaaktype = fetch_zaaktype(related_zaaktype_url)
+    informatieobjecttypen = get_informatieobjecttypen(related_zaaktype.catalogus)
     return DocumentSelectContext(
         documents=zaak_context.documents, informatieobjecttypen=informatieobjecttypen
     )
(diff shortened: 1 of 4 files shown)


backend/src/zac/core/camunda/select_documents/context.py
```diff
@@ -16,4 +18,10 @@
 )
 def get_context(task: Task) -> DocumentSelectContext:
     zaak_context = get_zaak_context(task, require_documents=True)
-    return DocumentSelectContext(documents=zaak_context.documents)
+    process_instance = get_process_instance(task.process_instance_id)
+    bijdrage_zaaktype_url = process_instance.get_variable("zaaktype")
+    bijdrage_zaaktype = fetch_zaaktype(bijdrage_zaaktype_url)
+    informatieobjecttypen = get_informatieobjecttypen(bijdrage_zaaktype.catalogus)
```
sergei-maertens, line 24: can we rename `bijdrage_zaak*` to `related_zaak*`? bijdrage is one particular type of related case (out of three options), and I'd like to avoid future confusion

sergei-maertens (changes requested):

backend/src/zac/core/camunda/select_documents/context.py
```diff
@@ -16,4 +18,10 @@
 )
 def get_context(task: Task) -> DocumentSelectContext:
     zaak_context = get_zaak_context(task, require_documents=True)
-    return DocumentSelectContext(documents=zaak_context.documents)
+    process_instance = get_process_instance(task.process_instance_id)
+    related_zaaktype_url = process_instance.get_variable("zaaktype")
+    related_zaaktype = fetch_zaaktype(related_zaaktype_url)
+    informatieobjecttypen = get_informatieobjecttypen(related_zaaktype.catalogus)
```
annashamray, line 24: Now you request all available document types from the catalog where the zaaktype is located. But they could not relate to this zaaktype, so if you try to relate a document object with this type to the zaak you'll have the 400 error from openzaak. Is it what you want? 
If you only need document types related to the particular zaaktype you need to use this endpoint - https://openzaak.utrechtproeftuin.nl/catalogi/api/v1/schema/#tag/zaaktype-informatieobjecttypen

annashamray (changes requested):

backend/src/zac/core/camunda/select_documents/context.py
damm89: Nope, that's not what I wanted! :-) Thanks!

annashamray (approved):

PR #286, opened 2021-04-14
Removed schema=None

damm89:


annashamray (commented):
I think the Activity endpoint can be tweaked a little bit:
`assignee` field now required an integer (which is a user id in DB). I think it's better to replace it with `User.username`. But I think it's better to discuss it with @kelvincy

damm89:
-> https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/290

PR #288, opened 2021-04-14
Feature/activities permissions

annashamray:
fixes https://github.com/GemeenteUtrecht/ZGW/issues/748

### DIFF
--- a/backend/src/zac/accounts/permission_loaders.py
+++ b/backend/src/zac/accounts/permission_loaders.py
@@ -32,7 +32,7 @@ def add_atomic_permission_to_user(
     atomic_permission = AtomicPermission.objects.create(
         object_type=object_type,
         object_url=object_url,
-        permission=zaken_inzien.name,
+        permission=permission_name,
     )
     user.atomic_permissions.add(atomic_permission)
     return atomic_permission
--- a/backend/src/zac/activities/api/permission_loaders.py
+++ b/backend/src/zac/activities/api/permission_loaders.py
@@ -1,0 +1,18 @@
+from typing import List
+
+from zac.accounts.models import AtomicPermission
+from zac.accounts.permission_loaders import add_atomic_permission_to_user
+
+from ..models import Activity
+from ..permissions import activiteiten_schrijven, activities_read
+
+
+def add_permissions_for_activity_assignee(activity: Activity) -> List[AtomicPermission]:
+    atomic_permissions = []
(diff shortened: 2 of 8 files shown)


backend/src/zac/accounts/permission_loaders.py
```diff
@@ -32,7 +32,7 @@ def add_atomic_permission_to_user(
     atomic_permission = AtomicPermission.objects.create(
         object_type=object_type,
         object_url=object_url,
-        permission=zaken_inzien.name,
+        permission=permission_name,
```
damm89, line 35: good catch(?)

backend/src/zac/activities/api/permission_loaders.py
```diff
@@ -0,0 +1,18 @@
...
+
+from ..models import Activity
+from ..permissions import activiteiten_schrijven, activities_read
+
+
+def add_permissions_for_activity_assignee(activity: Activity) -> List[AtomicPermission]:
+    atomic_permissions = []
+    for permission in [activities_read, activiteiten_schrijven]:
+        atomic_permission = add_atomic_permission_to_user(
+            activity.assignee, activity.zaak, permission_name=permission.name
```
damm89, line 14: Activity.zaak shouldn't be activity.url? In the current implementation the atomic permission relates to the zaak and not the activity.

backend/src/zac/api/permissions.py
```diff
@@ -79,6 +81,9 @@ class ObjectDefinitionBasePermission(DefinitionBasePermission):
     def get_object(self, request: Request, obj_url: str):
         raise NotImplementedError("This method must be implemented by a subclass")
 
+    def get_object_url(self, serializer=None) -> str:
+        return serializer.validated_data[self.object_attr]
```
damm89, line 85: if serializer = None this will break

damm89 (commented):
Looks, good. 1 remark and 1 question

backend/src/zac/activities/api/permission_loaders.py
annashamray: yeah, `object_type` for atomic permissions (and also blueprint permissions) can be `zaak` or `document` now, and we grant all permissions per zaak or per document

backend/src/zac/api/permissions.py
annashamray: Nice catch! I'll remove None as default value

damm89 (approved):

Issue #287, opened 2021-04-14
BFF endpoint adhoc activities


damm89:
Structure:

```
{
  finished: [],
  on_going: [],
}
```

Assignee field needs to be a User field (with User attributes such as name, etc).
Documents (are they needed?) field needs to be dowc read urls.

PR #290, opened 2021-04-15
Adapted activity endpoint for BFF

damm89:


backend/src/zac/activities/api/viewsets.py
```diff
@@ -1,26 +1,60 @@
...
+                location=OpenApiParameter.QUERY,
+            )
+        ],
+    ),
+    retrieve=extend_schema(summary=_("Retrieve activity")),
+    create=extend_schema(summary=_("Create activity")),
+    partial_update=extend_schema(
+        summary=_("Update activity"),
+        request=PatchActivitySerializer,
+        responses={201: ActivitySerializer},
```
damm89, line 35:
```suggestion
        responses={200: ActivitySerializer},
```

backend/src/zac/activities/api/serializers.py
```diff
@@ -43,3 +45,38 @@ class Meta:
...
+
+
+class PatchActivitySerializer(ActivitySerializer):
+    assignee = serializers.CharField(
+        label=_("assignee"),
+        help_text=_("User assigned to the activity."),
+        required=False,
+        allow_blank=True,
+        validators=(UserValidator(),),
+    )
```
annashamray, line 57: 1. I think `CharField` + `UserValidator` can be replaced with `SlugRelatedField`, there is an example here:
https://github.com/GemeenteUtrecht/zaakafhandelcomponent/blob/ee504f500d9fe460fb603f2c633cfd1e5a20e36e/backend/src/zac/accounts/api/serializers.py#L34

2. I don't like that `assignee` attribute has different shapes (string or object) depending on HTTP method. I think displaying username should be enough, and if not we have `users` BFF endpoint to extract user details

backend/src/zac/activities/api/serializers.py
```diff
@@ -43,3 +45,38 @@ class Meta:
...
+                    _(
+                        "Case URL is used for validation and can't be edited or left blank."
+                    )
+                )
+        return zaak_url
+
+    def update(self, instance, validated_data):
+        assignee = validated_data.pop("assignee", None)
+        if assignee:
+            validated_data["assignee"] = User.objects.get(username=assignee)
```
annashamray, line 80: Was it already checked in `UserValidator` ?

backend/src/zac/activities/api/serializers.py
```diff
@@ -43,3 +45,38 @@ class Meta:
...
+
+    def validate_zaak(self, zaak_url):
+        if hasattr(self, "instance") and isinstance(self.instance, Activity):
+            if self.instance.zaak != zaak_url:
+                raise serializers.ValidationError(
+                    _(
+                        "Case URL is used for validation and can't be edited or left blank."
+                    )
+                )
+        return zaak_url
```
annashamray, line 75: I think it would be nice to put this code into the separate validator in the utils and call it `ImmutableValidator` or something like this. And then we could reuse it for all fields which should not be changed in the endpoints

backend/src/zac/activities/api/viewsets.py
```diff
@@ -40,9 +74,27 @@ def filter_queryset(self, queryset):
...
+        serializer = self.get_serializer(
+            instance=instance, data=request.data, partial=True
+        )
+        serializer.is_valid(raise_exception=True)
+        self.perform_update(serializer)
+
+        if getattr(instance, "_prefetched_objects_cache", None):
+            # If 'prefetch_related' has been applied to a queryset, we need to
+            # forcibly invalidate the prefetch cache on the instance.
+            instance._prefetched_objects_cache = {}
```
annashamray, line 88: Could you please explain why you do it? And if you don't want django to cache prefetched objects why you use `prefetch_related` ?

backend/src/zac/activities/tests/test_api_response.py
```diff
@@ -0,0 +1,215 @@
...
+        }
+        endpoint = reverse(
+            "activities:activity-detail", kwargs={"pk": self.activity.pk}
+        )
+
+        # Mock zaak
+        m.get(self.zaak["url"], json=self.zaak)
+
+        # Put activity
+        response = self.client.put(endpoint, data=data)
```
annashamray, line 212: Why do you restrict the use of PUT method?

annashamray (changes requested):

annashamray:
@damm89  Is https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/287 a user story for this PR? 
If yes could you link them together, please?

backend/src/zac/activities/api/serializers.py
damm89: No, it can be empty

backend/src/zac/activities/api/serializers.py
damm89: TIL! :-)

backend/src/zac/activities/tests/test_api_response.py
damm89: There is no need to PUT right?

backend/src/zac/activities/api/viewsets.py
damm89: On an update they could be changed right (nested write serializer, for example)? Not really applicable in our case here, I guess, I just copied it from the update method from rest_framework. :-)

backend/src/zac/activities/tests/test_api_response.py
annashamray: .. yes :)

annashamray (approved):

Issue #295, opened 2021-04-20
Update authorization for ZaakDocuments endpoint


annashamray:
ZaakDocuments endpoint uses permissions both for zaken and for  documenten
Now it checks zaak permissions as usual (using `CanReadOrUpdateZaken` ) and after it filters documents based on document permissions with `filter_documenten_for_permissions` function. The function `filter_documenten_for_permissions` uses the old `PermissionsBackend` which is deprecated and should be removed. 

Todo:
* [x] remove the usage of `PermissionsBackend` from BFF endpoints
* [x] put as much code as possible to permissions to check documents.

damm89:
agree

PR #296, opened 2021-04-21
feature/bff-upload-edited-document

damm89:


backend/src/zac/core/api/bff_urls.py
```diff
@@ -35,13 +35,13 @@
...
-        ZaakDocumentsView.as_view(),
+        ListZaakDocumentsView.as_view(),
         name="zaak-documents",
     ),
     path(
         "cases/document",
-        CreateZaakDocumentView.as_view(),
-        name="add-document",
+        ZaakDocumentView.as_view(),
+        name="zaak-document",
```
annashamray, line 44: Now we have the same resource (ZaakDocument) which has GET, GET LIST, POST and PATH methods, but with different url structures.
I mean in the case of GET we need to specify `zaak.bronorganisatie` and `zaak.identificatie` in the path, and for PATCH and POST we don't need to do it despite the documents are created/updated for a particular zaak. 
It seems to break REST "uniform interface" principle.

annashamray (changes requested):

damm89 (commented):
.

backend/src/zac/core/api/bff_urls.py
damm89: The reason why is because I'm having some trouble with the permissions. 

Listing zaak documents checks the user's permissions regarding the `zaak` with `GetZaakMixin` and `CanReadZaken` which uses the `zaak` as `object_type`.
Creating a new document in the zaak checks the zaak permission through the `ZaakDefinitionPermission` and requires the zaak url to be provided in the serializer as `zaak`.
The permission for editing a document seems to me like it shouldn't have `zaak` as object_type but rather `document` and perhaps if a user can edit the zaak? In that case it would be better aligned.

annashamray: I agree that for creating and updating documents the permissions should be used for `document` object_type. I'm not sure if we should add permissions to edit the zaak here, but perhaps you are right. 

I like your document permissions. Let's leave it as it is for now and return to it when @sergei-maertens is around? 
The issue for it - https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/295

backend/src/zac/core/api/tests/test_manage_zaak_document.py
```diff
@@ -0,0 +1,504 @@
...
+
+        post_data = {
+            "informatieobjecttype": f"{CATALOGI_ROOT}informatieobjecttypen/d1b0512c-cdda-4779-b0bb-7ec1ee516e1b",
+            "zaak": f"{ZAKEN_ROOT}zaken/456",
+            "file": BytesIO(b"foobar"),
+        }
+
+        response = self.client.post(self.endpoint, post_data, format="multipart")
+
+        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```
annashamray, line 395: could you please also check that enkelvoudiginformatieobjecten mock and zaakinformatieobjecten mock were called
and also that the request body was as expected ?

annashamray (changes requested):

backend/src/zac/core/services.py
```diff
@@ -981,6 +988,19 @@ def update_document(url: str, data: dict, audit_line: str):
     return document
 
 
+def relate_document_to_zaak(document_url: str, zaak_url: str) -> Dict[str, str]:
```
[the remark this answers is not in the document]
damm89: Thank you.

sergei-maertens (approved):

PR #297, opened 2021-04-22
feature/bff-report-endpoint

damm89:


backend/src/zac/reports/api/views.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def has_permission(self):
+        # move logic from "reports:download" rule
+        report = self.get_object()
+        if not report:
+            return True
+
+        zaaktypen = get_zaaktypen(self.request.user)
+        identificaties = {zt.identificatie for zt in zaaktypen}
+        return set(report.zaaktypen).issubset(identificaties)
```
sergei-maertens, line 108: this belongs in a permission class

backend/src/zac/reports/api/views.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    def get(self, request, *args, **kwargs):
+        report = self.get_object()
+        ordering = ESOrderingFilter().get_ordering(self.request, self)
+        zaken, zaak_statuses, zaak_eigenschappen = get_zaken_details_for_export(
+            request.user, report, ordering=ordering
+        )
+
+        qs = []
```
sergei-maertens, line 77: Please give variables appropriate names. `qs` implies `queryset`, which implies a type of `models.QuerySet`, but this is a _list_ instead.

```suggestion
        results = []
```

backend/src/zac/reports/api/serializers.py
```diff
@@ -0,0 +1,23 @@
...
+            "name",
+        )
+
+
+class ReportDownloadSerializer(serializers.Serializer):
+    identificatie = serializers.CharField()
+    omschrijving = serializers.CharField()
+    startdatum = serializers.DateField()
+    status = serializers.CharField()
+    zaaktype_omschrijving = serializers.CharField()
```
sergei-maertens, line 20:
```suggestion
    zaaktype_omschrijving = serializers.CharField(source="zaaktype.omschrijving")
```
-> this way you can pass _objects_ in the view rather than doing the serialization manually.

backend/src/zac/reports/api/views.py
```diff
@@ -0,0 +1,108 @@
...
+            qs.append(
+                {
+                    "eigenschappen": eigenschappen,
+                    "identificatie": zaak.identificatie,
+                    "zaaktype_omschrijving": zaak.zaaktype.omschrijving,
+                    "startdatum": zaak.startdatum,
+                    "omschrijving": zaak.omschrijving,
+                    "status": zaak_statuses[zaak.url] if zaak.status else "",
+                }
+            )
```
sergei-maertens, line 95: you're essentially serializing manually here, you can prevent so much work by having (wrapper) objects that you can pass to the serializer. The essential information here is `zaak` and `eigenschappen`. This should be all you track in your result list, and pass that to the serializer. It's the serializer's job to access the (nested) attributes.

backend/src/zac/reports/export.py
```diff
@@ -51,8 +53,12 @@ def get_export_zaken(report: Report) -> List[Zaak]:
...
 
 
-def export_zaken(report: Report) -> tablib.Dataset:
-    zaken = get_export_zaken(report)
+def get_zaken_details_for_export(
+    user: User,
+    report: Report,
+    ordering: List[str] = ["startdatum", "registratiedatum", "identificatie"],
+) -> Tuple[List[Zaak], Dict[str, str], Dict[str, List[ZaakEigenschap]]]:
+    zaken = get_export_zaken(user, report, ordering=ordering)
```
sergei-maertens, line 61: I don't get why the tuple return type is needed, and having a list of `Zaak` instances is not sufficient?

1. https://github.com/GemeenteUtrecht/zaakafhandelcomponent/blob/main/backend/src/zgw/models/zrc.py#L14 -> you can already have `zaak.eigenschappen` be a list of `ZaakEigenschap` instances
2. `Zaak.status` should already resolve to a `Status` object

Imo this function should indeed fetch the data as efficiently as possible (which happens now), but the `Zaak` instances should be processed inside of this function.

That way, you can easily pass the `Zaak` instances to the serializers, which can extract _all_ the fields from just a `Zaak` instance, which massively simplifies the serializers/views.

backend/src/zac/reports/urls.py
```diff
@@ -1,10 +1,12 @@
...
 
+from .api import urls as api_urls
 from .views import DownloadReportView, ReportsListView
 
 app_name = "reports"
 
 urlpatterns = [
     path("", ReportsListView.as_view(), name="report-list"),
     path("download/<int:pk>/", DownloadReportView.as_view(), name="download"),
+    path("api/", include(api_urls)),
```
sergei-maertens, line 11: these result in ugly urls (`/reports/api/...`), please include the URLs in the correct place to have nice and logically structured URLs. `zac.api.urls` is the entrypoint for everything BFF related.

sergei-maertens (changes requested):

backend/src/zac/reports/api/views.py
damm89: Yeah, but I'd like to talk about this particular permission. Not sure where to put it.

backend/src/zac/reports/api/serializers.py
damm89: Thanks. Keep making this mistake... :(

backend/src/zac/reports/api/views.py
```diff
@@ -0,0 +1,86 @@
...
+        "omschrijving",
+        "startdatum",
+        "zaaktype.omschrijving",
+    )
+
+    def get(self, request, *args, **kwargs):
+        report = self.get_object()
+        ordering = ESOrderingFilter().get_ordering(self.request, self)
+        zaken = get_zaken_details_for_export(request.user, report, ordering=ordering)
+        page = self.paginate_queryset(zaken)
```
sergei-maertens, line 74: if we're supporting pagination, the `?page` querystring parameter should also be documented in the `@extend_schema` above.

backend/src/zac/reports/api/views.py
```diff
@@ -0,0 +1,86 @@
...
+        serializer = self.get_serializer(page, many=True)
+        return self.get_paginated_response(serializer.data)
+
+    def has_permission(self):
```
sergei-maertens, line 78: I don't see where this method is called though? There's also a test missing to check these permissions.

sergei-maertens (changes requested):

backend/src/zac/core/blueprints.py
```diff
@@ -119,3 +121,37 @@ def has_access(self, document: Document):
 
     def short_display(self):
         return f"{self.data['iotype_omschrijving']} ({self.data['max_va']})"
+
+
+class ReportBlueprint(Blueprint):
```
sergei-maertens, line 126: Please move this class to the `reports` app, the `core` app is meant for pure ZGW interaction, reports is extra functionality on top of that.

You really want to minimize imports between different Django apps to keep loose coupling.

sergei-maertens (changes requested):

PR #300, opened 2021-04-23
fix/camunda/select-documents

damm89:


backend/src/zac/core/camunda/select_documents/context.py
```diff
@@ -17,11 +21,16 @@
     DocumentSelectTaskSerializer,
 )
 def get_context(task: Task) -> DocumentSelectContext:
-    zaak_context = get_zaak_context(task, require_documents=True)
     process_instance = get_process_instance(task.process_instance_id)
+
+    zaak_url = process_instance.get_variable("hoofdZaakUrl")
```
sergei-maertens, line 26: For practical purposes I think we can use this for now, but it's super fragile because there is literally no guarantee this variable is present in a process instance.

sergei-maertens (approved):

PR #308, opened 2021-05-03
fix/find-document

damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/757.

### DIFF
--- a/backend/src/zac/core/services.py
+++ b/backend/src/zac/core/services.py
@@ -868,6 +868,8 @@ def find_document(
 
     # not in cache -> check it in all known DRCs
     drcs = Service.objects.filter(api_type=APITypes.drc)
+
+    result = None
     for drc in drcs:
         client = drc.build_client()
         results = get_paginated_results(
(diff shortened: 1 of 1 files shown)


backend/src/zac/core/services.py
```diff
@@ -886,9 +888,18 @@ def find_document(
...
             candidates = [result for result in results if result["versie"] == versie]
             if not candidates:
-                raise RuntimeError(
-                    f"Version '{versie}' for document does not seem to exist..."
-                )
+                # The DRC only returns the latest version and so the candidates
+                # will always be empty if the latest version isn't the requested version.
+                # In this case try to retrieve the document by using get_document.
+                try:
+                    url = results[0]["url"] + "?versie=%s" % versie
```
sergei-maertens, line 895: that's some ugly string manipulation to build URLs and not future proof if at some point `versie` could contain URL-unsafe characters.

We already have [furl](https://github.com/gruns/furl) in our requirements, let's make use of it too ;)

sergei-maertens (changes requested):

damm89:
See https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/310

PR #310, opened 2021-05-04
Issue/document cache rework

damm89:


backend/src/zac/contrib/dowc/views.py
```diff
@@ -39,15 +40,16 @@ def _cast(value: Optional[Any], type_: type) -> Any:
     ],
 )
 class OpenDowcView(APIView):
-    authentication_classes = (authentication.SessionAuthentication,)
+    # authentication_classes = (authentication.SessionAuthentication,)
```
damm89, line 43:
```suggestion
    authentication_classes = (authentication.SessionAuthentication,)
```

backend/src/zac/contrib/dowc/views.py
```diff
@@ -87,7 +88,7 @@ def delete(self, request, dowc_uuid):
         serializer.is_valid(raise_exception=True)
         data = patch_and_destroy_doc(request.user, serializer.validated_data["uuid"])
 
-        # Invalidate cache if valid response
+        # Refresh cache state if valid response
```
damm89, line 91:
```suggestion
        # Invalidate cache if valid response
```

backend/src/zac/core/services.py
```diff
@@ -57,12 +58,20 @@ def _client_from_object(obj):
 
 
 async def fetch(session: aiohttp.ClientSession, url: str):
+    """
+    DEPRECATED/UNUSED
```
sergei-maertens, line 62: if these are not used anymore, why not just rip it out from the codebase?

backend/src/zac/core/views/documents.py
```diff
@@ -29,7 +30,8 @@ def has_permission(self):
 
     def get_object(self) -> Document:
         versie = _cast(self.request.GET.get("versie", None), int)
-        self.document = find_document(versie=versie, **self.kwargs)
+        document = find_document(versie=versie, **self.kwargs)
+        self.document = factory(Document, document)
```
sergei-maertens, line 34: I don't like how inconsistent the Python API is now - the vision was always that `core/services.py` would return python domain objects ready for use in other layers, now the creation of domain objects is delegated to those layers instead which introduces a lot of boilerplate code.

sergei-maertens (changes requested):

backend/src/zac/core/services.py
damm89: Sure - np.

backend/src/zac/core/services.py
```diff
@@ -847,27 +906,24 @@ def get_documenten(
...
-        cache_key = f"document:{document.bronorganisatie}:{document.identificatie}"
-        cache.set(cache_key, document, timeout=AN_HOUR / 2)
-
     return documenten, gone
 
 
-@cache_result(
-    "document:{bronorganisatie}:{identificatie}:{versie}", timeout=AN_HOUR / 2
-)
+@cache_result("document:{bronorganisatie}:{identificatie}:{versie}")
```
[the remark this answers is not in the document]
damm89: Shouldnt be :-)

backend/src/zac/core/services.py
```diff
@@ -839,85 +866,94 @@ def get_documenten(
...
-)
 def find_document(
     bronorganisatie: str, identificatie: str, versie: Optional[int] = None
 ) -> Document:
     """
     Find the document uniquely identified by bronorganisatie and identificatie.
+
     """
-    query = {"bronorganisatie": bronorganisatie, "identificatie": identificatie}
+    cache_key = "document:{bronorganisatie}:{identificatie}:{versie}"
```
sergei-maertens, line 885:
```suggestion
    cache_key = f"document:{bronorganisatie}:{identificatie}:{versie}"
```

and the fact the build didn't choke on this indicates that there's a missing test for the caching behaviour of this function :P

sergei-maertens (changes requested):
One bug spotted :-)

sergei-maertens (approved):

PR #311, opened 2021-05-05
feature/dowc-error-relay

damm89:


backend/src/zac/contrib/dowc/api.py
```diff
@@ -43,29 +43,29 @@ def create_doc(
...
+    try:  # First try to create the object in dowc
+        response = client.create("documenten", data=data)
+        status_code = status.HTTP_201_CREATED
+
+    except ClientError as err:
+        if err.args == (
+            None,
+        ):  # Requesting user may already have created an object dowc
+            try:
+                response = client.list("documenten", data=data)
```
damm89, line 61:
```suggestion
            try:
                # We can fetch the first from the list because the
                # client will have raised an exception
                # if the status isn't as expected.
                response = client.list("documenten", data=data)[0]
```

backend/src/zac/contrib/dowc/views.py
```diff
@@ -39,15 +40,16 @@ def _cast(value: Optional[Any], type_: type) -> Any:
     ],
 )
 class OpenDowcView(APIView):
-    authentication_classes = (authentication.SessionAuthentication,)
+    # authentication_classes = (authentication.SessionAuthentication,)
```
damm89, line 43:
```suggestion
    authentication_classes = (authentication.SessionAuthentication,)
```

backend/src/zac/contrib/dowc/views.py
```diff
@@ -39,15 +40,16 @@ def _cast(value: Optional[Any], type_: type) -> Any:
...
     permission_classes = (permissions.IsAuthenticated & CanOpenDocuments,)
     document = None
     serializer_class = DowcResponseSerializer
 
     def get_object(self, bronorganisatie: str, identificatie: str) -> Document:
         if not self.document:
             versie = _cast(self.request.GET.get("versie", None), int)
-            self.document = find_document(bronorganisatie, identificatie, versie=versie)
+            document = find_document(bronorganisatie, identificatie, versie=versie)
+            self.document = factory(Document, document)
```
damm89, line 52:
```suggestion
            self.document = find_document(bronorganisatie, identificatie, versie=versie)
```

backend/src/zac/contrib/dowc/views.py
```diff
@@ -87,7 +92,7 @@ def delete(self, request, dowc_uuid):
         serializer.is_valid(raise_exception=True)
         data = patch_and_destroy_doc(request.user, serializer.validated_data["uuid"])
 
-        # Invalidate cache if valid response
+        # Refresh cache state if valid response
```
damm89, line 95:
```suggestion
        # Invalidate cache if valid response
```

backend/src/zac/contrib/dowc/api.py
```diff
@@ -43,29 +43,32 @@ def create_doc(
...
+
+    data = {"drc_url": drc_url, "purpose": purpose, "info_url": referer}
+    try:  # First try to create the object in dowc
+        response = client.create("documenten", data=data)
+        status_code = status.HTTP_201_CREATED
+
+    except ClientError as err:
+        if err.args == (
+            None,
+        ):  # Requesting user may already have created an object dowc
```
sergei-maertens, line 59: not sure where/when this happens, since in the zds-client code it appears this only happens if the response cannot be parsed as json: https://github.com/VNG-Realisatie/gemma-zds-client/blob/master/zds_client/client.py#L215

backend/src/zac/contrib/dowc/api.py
```diff
@@ -43,29 +43,32 @@ def create_doc(
...
+            try:
+                # We can fetch the first from the list because the
+                # client will have raised an exception
+                # if the status isn't as expected.
```
sergei-maertens, line 63: ah I see, this is happening because the `create` call in DOWC does not always lead to a HTTP 201, if the thing already exists it will reply with a HTTP 200 :thinking: 

maybe we could change zds-client to accept a list of `accepted_status` to provide multiple okay statuses, and expose that through the `.create(...)` etc. methods so that you get more control about it. This code-block would then essentially become:

```py
response = client.create("documenten", data=data, expected_status=[201, 200])
```
you do lose the information on if the object was created or not (200 vs. 201), so I'm not too sure about this approach.

backend/src/zac/contrib/dowc/api.py
```diff
@@ -43,29 +43,32 @@ def create_doc(
...
+                response = client.list("documenten", data=data)[0]
+                status_code = status.HTTP_200_OK
+            except ClientError as err:  # Relay error
+                return err.args[0], status.HTTP_400_BAD_REQUEST
+        else:  # Object might exist and is owned by a different user
+            return err.args[0], status.HTTP_400_BAD_REQUEST
```
sergei-maertens, line 69: this I would turn into a custom exception instead:

```py
raise DOWCCreateError(err.args[0]) from err
```

instead of having different return types.

backend/src/zac/contrib/dowc/views.py
```diff
@@ -57,6 +57,9 @@ def post(self, request, bronorganisatie, identificatie, purpose):
         document = self.get_object(bronorganisatie, identificatie)
         referer = request.headers.get("referer", "")
         response, status_code = create_doc(request.user, document, purpose, referer)
+        if status_code == 400:
+            return Response(response, status=status_code)
```
sergei-maertens, line 61: This would then become:

```suggestion
        try:
            response, status_code = create_doc(request.user, document, purpose, referer)
        except DOWCCreateError as err:
            raise exceptions.ValidationError(err.args[0])
```

sergei-maertens (changes requested):

backend/src/zac/contrib/dowc/api.py
```diff
@@ -1,4 +1,4 @@
-from typing import Dict, Optional, Tuple
+from typing import Any, Dict, Tuple, Union
```
sergei-maertens, line 1: betrayed by the linter again I think, `Any` & `Union` are unused.

sergei-maertens (approved):

PR #317, opened 2021-05-14
feature/medewerker-name

damm89:


backend/src/zac/core/services.py
```diff
@@ -666,6 +671,54 @@ def fetch_rol(rol_url: str) -> Rol:
...
+
+
+def update_medewerker_identificatie_rol(zaak: Zaak) -> Optional[List[Rol]]:
+    rollen = get_rollen(zaak)
+
+    from .api.serializers import UpdateRolSerializer
+
+    new_rollen = []
+    for rol in rollen:
+        if rol.betrokkene_type.lower() == RolTypes.medewerker.lower():
```
sergei-maertens, line 694: there's no point in doing a `.lower()` call, the types are an enum in the API spec and the API validates these values.

sergei-maertens (changes requested):
small remark

sergei-maertens (approved):

PR #323, opened 2021-05-19
fix/cache-rollen

damm89:


backend/src/zac/core/cache.py
```diff
@@ -71,11 +71,7 @@ def invalidate_document_cache(document: Document):
...
 
-def invalidate_rollen_cache(zaak: Zaak, rollen: Optional[List[str]] = None):
+def invalidate_rollen_cache(zaak: Zaak):
     _cache = caches["request"]
     if _cache:
-        _cache.delete(f"report:{zaak.url}")
-
-    if rollen:
-        cache_keys = [f"rol:{rol}" for rol in rollen]
-        cache.delete_many(cache_keys)
```
annashamray, line 81: this function is also called when the notification about destroying the rol is received. I think in this case caches for particular rollen `"rol:{rol}"` should also be deleted

annashamray (changes requested):

backend/src/zac/core/cache.py
damm89: If the rol gets destroyed, its cache will be invalidated by the function that destroys it. Should it be moved?

annashamray: the cache would be cleared if the rol is deleted in the ZAC. But what if it was deleted somewhere else? (like in OpenZaak admin). In this case we will know about it only via notifications

damm89: You're right, thanks.

sergei-maertens (approved):
Looks fine, just fix the build before merging :-)

sergei-maertens (approved):

Issue #320, opened 2021-05-17
Feature save search queries as reports


damm89:
It makes sense to combine search and report functionality as they both get their information from elasticsearch and it allows the searching user to save their search for reporting.

The questions that come forward are (and current state of affairs as of June 1st 2021):

**Functional/UI/UX**
- What does a query consist of? (i.e. which columns can users filter & search on and which columns can they show?)
> - For now a query is a result from the search in searches.py. By using the fields key in the post payload, users can select which fields they will see. Fields will ALWAYS consist of at least `identificatie`.
- How do we save queries as reports? 
> - We save the search query as given to search in searches.py and query the ES with this search query.
- How do we let the FE sort? 
> - Through the ESOrderingFilter.

**Permissions**
- Who can save their queries as reports? 
> - Any end user for now.
- Who can see those reports? 
> - Those who have the permission to the zaaktype.
- Who can manage reports? 
> - Django admins.

damm89:
### How do we let the FE sort?
Currently the data can be sorted on the columns of the ES `Document`, even if they are nested objects.

ES requires dot notation for nested fields -> `zaaktype.omschrijving` for example.
Django uses double underscore for nested fields -> `zaaktype__omschrijving` for example.
The current serializer has a 'custom' variable that is a serializers.CharField which grabs the source (`zaaktype.omschrijving`) but is assigned to `zaaktype_omschrijving` which gets returned as `zaaktypeOmschrijving` by our current renderer.

I need some help figuring out the most comfortable alignment strategy for all parties involved. :-)

damm89:
### Plan de campagne

**End goal**
Users will be able to create a report by saving their ES search query on the /ui/zaken page and are able to retrieve and destroy that report.

**Possible Implementation**
- [ ] Add ES version(s) of ZaakDetailSerializer (and nested serializers) where no fields are required, but _can_ be returned.
- [ ] Filter out all fields as last step before pagination in view rather than in ES.
- [ ] Add `fields` multiple choice field to SearchSerializer in https://github.com/GemeenteUtrecht/zaakafhandelcomponent/blob/43eea0699363a7447523a6106519a38a95940126/backend/src/zac/elasticsearch/drf_api/serializers.py#L26.
> - [ ] Valid choices of multiple choice field are fields as found ES-modified ZaakDetailSerializer

[...]

PR #328, opened 2021-06-01
feature/search-reports

damm89:


damm89:
Closes https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/320

backend/src/zac/elasticsearch/drf_api/views.py
```diff
@@ -44,53 +55,171 @@ def get(self, request: Request) -> Response:
...
     serializer_class = SearchSerializer
-    search_document = ZaakDocument
-    ordering = (
-        "-identificatie",
-        "-startdatum",
-        "-registratiedatum",
-    )
+
+    def get_queryset(self, **kwargs):
+        return self.perform_search(kwargs)
```
sergei-maertens, line 102: This causes the following drf-spectacular warning:

```
Warning #2: SearchView: failed to obtain model through view's queryset due to raised exception. Prevent this either by setting "queryset = Model.objects.none()" on the view, having an empty fallback in get_queryset() or by using @extend_schema. (Exception: Cannot cast AnonymousUser to int. Are you trying to use it in place of User?)
```

I would really, really like to avoid introducing _more_ warnings for schema generation. This is also semantically incorrect, as `get_queryset` implies a `models.QuerySet` instance which this is not.

It makes more sense to not use `GenericAPIView` functionality here, but rather implement `def post(...)` accordingly and make the relevant calls. It does not fit the `ModelViewSet` model, since no models are involved.

API documentation should then be done through `@extend_schema` options. I'm not sure _why_ the drf-yasg documentation needs the model, but probably we can avoid it trying to retrieve it through some explicit schema documentation.

backend/src/zac/elasticsearch/drf_api/pagination.py
```diff
@@ -0,0 +1,21 @@
...
+
+from rest_framework.response import Response
+
+from zac.core.api.pagination import BffPagination
+
+from ..documents import ZaakDocument
+from .utils import get_document_fields, get_document_properties
+
+
+class ESPagination(BffPagination):
```
sergei-maertens, line 11: This will require a drf-spectacular plugin to modify the schema, since looking at http://localhost:8000/api/docs/#operation/search_reports_results now, the `fields` response key is not documented. You should be able to override https://github.com/encode/django-rest-framework/blob/master/rest_framework/pagination.py#L340, which is called here: https://github.com/tfranzel/drf-spectacular/blob/master/drf_spectacular/openapi.py#L394. We already have some extensions in `src/zac/api/drf_spectacular` so you can use that for inspiration.

backend/src/zac/elasticsearch/drf_api/views.py
```diff
@@ -44,53 +55,171 @@ def get(self, request: Request) -> Response:
...
+        # only allow reports where the zaaktypen are a sub-set of the accessible zaaktypen
+        # for this particular user
+        allowed_report_ids = []
+        for report in qs:
+            search_results = self.perform_search(report.query)
+            zaaktypen = {result.zaaktype.url for result in search_results}
+            if zaaktypen.issubset(allowed_zt_urls):
+                allowed_report_ids.append(report.id)
+
+        return qs.filter(id__in=allowed_report_ids)
```
sergei-maertens, line 164: This looks a bit illogical - if you have a saved report (say, it's based on zaak-omschrijving). One day you can have matches that fit within your allowed zaaktypen, and the next day a new zaaktype was created with zaken that match but you don't have access permissions on the zaaktype, so now the entire report is no longer usable? Do I interpret that code right?

If that's the case, it'd make more sense that the report is accessible but the results are filtered based on your permissions (or even better - the report would be taken as a base query, and the permissions are consulted to apply the zaaktypen as extra constraints, which gives you the most efficient ES query).

What you _can_ do, is check if the query includes a set of zaaktypen, and check those zaaktypen for being a subset of the zaaktypen you have access to. Then you can reject permission, and not even hit elastic search because you know up-front that there will/may be invalid results. The current approach with multiple ES-hits seems a sub-optimal approach w/r to performance and expected results.

I realize that's a complex thing to build, so perhaps we need to discuss this with the business.

backend/src/zac/elasticsearch/drf_api/views.py
```diff
@@ -44,53 +55,171 @@ def get(self, request: Request) -> Response:
...
+    def partial_update(self, request, *args, **kwargs):
+        return super().partial_update(request, *args, **kwargs)
+
+    @extend_schema(summary=_("Retrieve a search report"))
+    def retrieve(self, request, *args, **kwargs):
+        return super().retrieve(request, *args, **kwargs)
+
+    @extend_schema(summary=_("Update a search report"))
+    def update(self, request, *args, **kwargs):
+        return super().update(request, *args, **kwargs)
```
sergei-maertens, line 190: Use `@extend_view_schema(foo=extend_schema(...), bar=extend_schema(...))` on the viewset instead, where `foo, bar` are the method names. It's a class decorator that decorates the underlying methods.

backend/src/zac/elasticsearch/drf_api/views.py
```diff
@@ -44,53 +55,171 @@ def get(self, request: Request) -> Response:
...
+
+
+class SearchReportDetailView(ESGenericAPIView):
+    parser_classes = (IgnoreCamelCaseJSONParser,)
+    permission_classes = (IsAuthenticated & CanDownloadSearchReports,)
+    queryset = SearchReport.objects.all()
+    serializer_class = ZaakDocumentSerializer
```
sergei-maertens, line 197: This should be a detail action on the viewset, since it's directly tied to the same resource (`SearchReport`), something like:

```py
@action(detail=True, permission_classes=(...))
def results(self, request, *args, **kwargs):
    ...
```

with `rest_framework.decorators.action`. That will then show up properly in the API docs too & URL routing is handled via the viewset/router, see earlier comment in the `urls.py` file

backend/src/zac/elasticsearch/tests/factories.py
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+
+class SearchReportFactory(factory.django.DjangoModelFactory):
+    name = factory.Sequence(lambda n: "search-report-%d" % n)
+
+    class Meta:
+        model = "elasticsearch.SearchReport"
+
+    query = factory.LazyFunction(get_random_search_query)
```
sergei-maertens, line 25: please follow django's model convention in factories and put this field before the `class Meta:` declaration.

sergei-maertens (changes requested):
Additionally, I think the `tablib` dependency is now unused and can be deleted!

backend/src/zac/elasticsearch/drf_api/views.py
damm89: I did something like this before but you would need to update those zaaktypen in the database of the model whenever new zaaktypen get added to the report results.

backend/src/zac/elasticsearch/drf_api/urls.py
```diff
@@ -1,8 +1,19 @@
...
+router = SimpleRouter()
+router.register(r"reports", SearchReportViewSet)
 
 urlpatterns = [
+    path("", include(router.urls)),
+    path(
+        "reports/<int:pk>/results",
+        SearchReportDetailView.as_view(),
+        name="searchreport-results",
+    ),
```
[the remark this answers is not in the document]
damm89: Works like a beast!

backend/src/zac/elasticsearch/drf_api/pagination.py
damm89: Thanks, I was looking for this functionality.

backend/src/zac/elasticsearch/drf_api/views.py
damm89: I guess the question that needs answering is: What is a report?

If reports should be reproducible for every user that access to that particular report then the current implementation stands. 

Otherwise, if reports should show results that fit the query and it's okay if the same report produces different results for users given their particular set of permissions, then we should filter out parts of the results that they don't have access to.

I just assumed the safest route would be the first definition, but you're right, this is something the team should define.

sergei-maertens: The route to apply additional filtering based on permissions was picked as the desired one (during standup today)

sergei-maertens (approved):
nice!

PR #335, opened 2021-06-10
Issue/#781

svenvandescheur:


svenvandescheur:
Dynamisch maken op basis van open forms prototype api.

svenvandescheur:
@sergei-maertens wil jij een blik werpen op de backend?

backend/src/zac/forms/api/views.py
```diff
@@ -0,0 +1,38 @@
+from drf_spectacular.utils import extend_schema
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from ..client import OpenFormsClient
+
+
+class OpenFormsFormListView(APIView):
```
sergei-maertens, line 9: let's avoid confusion with the real Open Forms and express clearly this is for the prototype

```suggestion
class OpenFormsProtoFormListView(APIView):
```

backend/src/zac/forms/api/views.py
```diff
@@ -0,0 +1,38 @@
...
+                        "layouts": {
+                            "type": "array",
+                            "items": {
+                                "type": "string",
+                            },
+                        },
+                    },
+                },
+            }
+        },
```
sergei-maertens, line 33: Hrm, normally this would be introspected from serializers. It's a shame the protoype doesn't expose API docs.

I'd still rather use serializers, because when we switch to the actual Open Forms, this API documentation of the BFF will probably go out of sync with the actual implementation. Using serializers from the start will require us to update the serializers and then the docs are automatically updated as well.

Can we define:

```py
class OpenFormsProtoFormSerializer(serializers.Serializer):
    id = serializers.IntegerField()
    name = serializers.CharField()
    slug = serializers.SlugField()
    layouts = serializers.ListField(child=serializers.CharField())
```

& then use it as `serializer_class` on the view

backend/src/zac/forms/api/views.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    def get(self, request, format=None):
+        client = OpenFormsClient()
+        definitions = client.get_forms()
+        return Response(definitions)
```
sergei-maertens, line 38: with the suggested above changes:

```suggestion
        definitions = client.get_forms()
        serializer = self.get_serializer(instance=definitions)
        return Response(serializer.data)
```

& define

```py
def get_serializer(self, **kwargs):
    return OpenFormsProtoFormSerializer(many=True, **kwargs)
```

this ensures drf-spectacular correctly picks up the `many=True`.

backend/src/zac/forms/tests/test_api.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class ResponseTests(APITestCase):
+    def setUp(self):
+        user = SuperUserFactory.create()
+        self.client.force_authenticate(user=user)
+
+    @patch("zac.forms.client.OpenFormsClient.__init__")
+    @patch("zac.forms.client.OpenFormsClient.get_forms")
+    def test_view_report(self, mock_get_forms, mock_init):
```
sergei-maertens, line 18: bad copy-paste test name

sergei-maertens (changes requested):

frontend/zac-ui/libs/features/open-forms/src/lib/features-open-forms.component.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    get bodyData(): Array<RowData> {
+        return this.forms.map(form => ({
+                cellData: {
+                    'Naam': form.name,
+                    'URL': {
+                        type: "link",
+                        label: `https://formulieren-test.utrechtproeftuin.nl/forms/${form.slug}/`,
+                        url: `https://formulieren-test.utrechtproeftuin.nl/forms/${form.slug}/`,
```
kelvincy, line 39: de url hiervan kan ook `formulieren.utrechtproeftuin.nl/forms` zijn, zonder `test`

kelvincy (changes requested):

backend/src/zac/forms/api/views.py
svenvandescheur: As discussed: use generic terminology instead.

backend/src/zac/forms/api/views.py
svenvandescheur: Much better indeed, I've added the Serializer.

backend/src/zac/forms/tests/test_api.py
svenvandescheur: whoops

frontend/zac-ui/libs/features/open-forms/src/lib/features-open-forms.component.ts
svenvandescheur: As discussed: create a development environment check function in utils based on the window location for now. I'll add an issue to investigate/implement a more solid check.

kelvincy (approved):

sergei-maertens (approved):

PR #336, opened 2021-06-10
:sparkles: #776 - As a zac user i want to have limited choice between…

svenvandescheur:
… aard of a relation #776


svenvandescheur:
There are some concirns whether we should filter this. Lets discuss further.

svenvandescheur:
As discussed during standup: we don't want this change for now.

PR #345, opened 2021-06-15
:sparkles: #780 - Add alfresco link to ZAC nav.

svenvandescheur:


frontend/zac-ui/apps/zac-ui/src/app/constants/menu.ts
```diff
@@ -30,6 +30,12 @@ const bottomMenuItems: MenuItem[] = [
   //   to: '/accounts/login/?next=/ui/',
   //   external: true
   // },
+  {
+    icon: 'link',
+    label: 'Alfresco',
+    to: 'https://alfresco-oz.utrechtproeftuin.nl/',
```
sergei-maertens, line 36: we really need to find a solution to avoid hardcoding links, as this will (in the near future) be hosted on the `*.aks.utrechtproeftuin.nl` cluster with a different domain name

kelvincy (approved):

frontend/zac-ui/apps/zac-ui/src/app/constants/menu.ts
svenvandescheur: Yes, see also: https://github.com/GemeenteUtrecht/ZGW/issues/792. I think right now we always do the same build as we run in containers. Maybe add a mechanism to pass environment variables/settings files?

svenvandescheur: I this something you consider blocking for this PR?

sergei-maertens: with static builds like Angular & React SPA's configuration through environment variables is not easy, since build artifacts need to be changed at container runtime. And I'm not looking to build different images for different environments.

Wouldn't say it's blocking now, but it we need a structural solution asap.

PR #355, opened 2021-06-21
:sparkles: #792 - Implement a mechanism for accurate runtime environm…

svenvandescheur:
…ent information (dev/production etc.)

This implements the environment files, the .prod versions are used by the docker built. For ingress-dev the regular one is used which contains development settings.

This does not implement a bridge API for settings, as this is something that might need to have more clarification.


sergei-maertens:
Oh hold - the whole image building approach needs to be discussed

sergei-maertens:
@svenvandescheur can this be closed? I think we all agreed that this would not be the ideal way forward and since it's pending for so long, there's obviously no urgency either.

Feel free to re-open if this is still relevant.

PR #359, opened 2021-06-25
remove excessive dates from permission models

annashamray:
In practice start and end dates for permissions make sense only in relation to users. Therefore:
* dates from `AtomicPermission` model are moved to `UserAtomicPermission` model
* dates from `BlueprintPermission` model are removed, since they duplicated the function of `UserAuthorizationProfile` model dates
* removing dates from permission models make it possible to remove duplicated permissions and ensure that the permission objects are unique.


damm89 (approved):
I like it! Improves the way permissions, users and objects are related to each other.

backend/src/zac/accounts/models.py
```diff
@@ -229,22 +230,13 @@ class AtomicPermission(models.Model):
...
-        null=True,
-        help_text=_("End date of the permission"),
-    )
+
     objects = AtomicPermissionQuerySet.as_manager()
 
     class Meta:
         verbose_name = _("atomic permission")
         verbose_name_plural = _("atomic permissions")
+        unique_together = ("permission", "object_url")
```
sergei-maertens, line 239: For the future: what happens if a user has a permission, it then expires, and say, half a year later they get the same permission again?

sergei-maertens (approved):
Nice refactor

PR #361, opened 2021-06-29
Fix/fe execute url 337

kelvincy:


svenvandescheur (approved):
Should we start adding documentation to new components?

kelvincy:
> Should we start adding documentation to new components?

will do!

PR #362, opened 2021-06-29
document authorization concept

annashamray:


svenvandescheur (approved):
Turned out to of good assistance, so works as intended 👍

damm89 (approved):
Nicely put together. I have two questions:

1) Don't we also have "Report"/"SearchReport" permissions or did I implement it wrong and will it be changed sometime?
2) I think the arrows in the picture of the relationships point the wrong way, unless there's some convention I'm unaware of. For example, the accounts_blueprintpermission and accounts_authorizationprofile are required to create accounts_authorizationprofile_blueprint_permissions, right?

sergei-maertens:
@damm89 don't think it's wrong, but rather incomplete - let's update the documentation with the extra permissions!

sergei-maertens (approved):

sergei-maertens (changes requested):
Please update the relations image (last one) with the refactored dates :grimacing:

kelvincy (approved):

annashamray:
@sergei-maertens 
ER diagram is updated

@damm89 
1. I've missed that the report permissions were created. I've added them to the doc now.
2. I use PyCharm to generate ER diagram, if it confuses more than it helps I can just remove it

damm89:
> I've missed that the report permissions were created. I've added them to the doc now.

No, no need to remove. They definitely help. :+1:

sergei-maertens (approved):

PR #365, opened 2021-07-02
Issue/#817

svenvandescheur:
angular upgrade
angular-material upgrade
table.component.ts rewrite om gebruik te maken van mat-table
user-permissions.component.ts op basis van table.component.ts


frontend/zac-ui/libs/features/zaak-detail/src/lib/user-permissions/user-permissions.component.ts
```diff
@@ -0,0 +1,138 @@
...
+                    delete: {
+                        type: 'button',
+                        label: 'Verwijderen',
+                        value: permission,
+                    }
+                },
+                nestedTableData: new Table(['Reden', 'Commentaar'], [{
+                    cellData: {
+                        reason: permission.reason,
+                        comment: permission["comment?"]
```
kelvincy, line 100: The question mark can be removed. The value is not showing up.

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
```diff
@@ -1,79 +1,100 @@
...
+    <ng-container matColumnDef="_expandedIndent">
+        <td class="table__cell" mat-cell *matCellDef="let dataRow"></td>
+    </ng-container>
 
-      <!-- Table -->
-      <tr *ngIf="expandable && element.nestedTableData" class="child-row nested-table-row">
-        <td colspan="10">
-          <gu-table [tableData]="element.nestedTableData"
-                    (buttonOutput)="handleNestedButtonClick($event)"></gu-table>
+    <ng-container matColumnDef="_expandedDetail">
```
kelvincy, line 66: This is always visible, even when it's empty

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
```diff
@@ -1,79 +1,100 @@
...
+                <gu-table [table]="dataRow._nestedTableData" (buttonOutput)="onNestedButtonClick($event)"></gu-table>
+            </div>
         </td>
-      </tr>
-    </ng-template>
-  </tbody>
+    </ng-container>
+
+    <!-- Header row -->
+    <tr mat-header-row *matHeaderRowDef="displayedColumns"></tr>
```
kelvincy, line 78: This is always visible, even when it's empty

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
```diff
@@ -1,79 +1,100 @@
...
-        [attr.role]="element.clickOutput || (expandable && element.expandData) ? 'button' : ''"
-      >
-        <!-- Expand icon -->
-        <ng-container>
-          <td *ngIf="(expandable && element.expandData) || (expandable && element.nestedTableData)" class="table-expand">
-            <mat-icon class="arrow" aria-hidden="false" aria-label="expand">keyboard_arrow_right</mat-icon>
-          </td>
-          <td *ngIf="(expandable && (!element.expandData  && !element.nestedTableData))"></td>
-        </ng-container>
+<table *ngIf="dataSource" class="mat-elevation-z8 table" [class.table--wrap]="wrap" mat-table matSort
```
kelvincy, line 1: I find this elevation a bit heavy

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.scss
```diff
@@ -2,96 +2,88 @@
 
 :host {
   display: block;
-  width: 100%;
-  overflow-x: auto;
-  overflow-y: hidden;
 }
 
 .table {
```
kelvincy, line 7: `mat-table` is using a different `font-family` than the rest of the body

frontend/zac-ui/libs/features/zaak-detail/src/lib/user-permissions/user-permissions.component.ts
```diff
@@ -0,0 +1,138 @@
...
+                'Acties',
+            ], bodyData);
+    }
+
+    //
+    // Events.
+    //
+
+    /**
+     * TODO: VALIDATE, DOES THIS WORK???
```
kelvincy, line 122: Works as expected :)

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.service.ts
```diff
@@ -0,0 +1,136 @@
...
+   * @return {string[]}
+   */
+  getDisplayedColumnNames(columns: Column[], expandable = false): string[] {
+    const uiColumns = expandable ? [{name: '_expandableToggle', label: ''}] : [];
+    return [...uiColumns, ...columns].map(c => c.name);
+  }
+
+  //
+  // Legacy data transforms.
+  // TODO: Consider migrating to a new table format?
```
kelvincy, line 102: Yes, I think we should allow a free form for the table cell instead of defining it with a type every time.

kelvincy (changes requested):
Nice job refactoring the table component. Displaying and removing the user permissions also works fine!

A few comments in general:
- The "expand" arrow on the left side of the table can get cramped up at some resolutions. Maybe add some padding?
- In the component `gu-documenten` the button "Bewerken" will change on "Bewerkingen opslaan" after opening the editable file. This button change is not happening anymore with the new table. Can you have a look at that?

frontend/zac-ui/libs/features/zaak-detail/src/lib/user-permissions/user-permissions.component.ts
svenvandescheur: :)

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
svenvandescheur: It seems to be the default, what value would you suggest?

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
svenvandescheur: Table columns are always specified so don't need a check on it being present. When te table has no (loaded) data (yet) I think the headers should still show providing some context about data that might become available.

Do you think it's better to not show headers if there is no data (loaded/yet)?

frontend/zac-ui/libs/features/zaak-detail/src/lib/user-permissions/user-permissions.component.ts
svenvandescheur: fixed

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
svenvandescheur: It's visible to the browser, not the user. It might be better to hide it from the DOM as well (although this has performance impact) but this seems to be problematic for some reason (*ngIf doesn't respond properly, possibly due to mat-table implementation). I can investigate this. but given performance doesn't really seem to be worth the effort. Please let me now if this is a dealbreaker for you.

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.scss
svenvandescheur: I tried to divert as little as possible from factory angular material styling as this is where we're migrating to. I think it would be a better idea to migrate the rest of the body as well.

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.service.ts
svenvandescheur: I think this is something out of scope for now.

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
kelvincy: I agree. Just leave it as is then

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
kelvincy: Each table row has a bit thicker bottom line because of some empty DOM elements. I think it's better to remove it if it's not used.

frontend/zac-ui/libs/shared/ui/components/src/lib/components/table/table.component.html
kelvincy: I suggest `mat-elevation-z1` for now.

kelvincy (approved):

PR #367, opened 2021-07-06
Feature/email lookup

damm89:


damm89:
Related to https://github.com/GemeenteUtrecht/bptl/pull/151
Fixes:

sergei-maertens (approved):

sergei-maertens (changes requested):
@damm89 can you bring back the `include` and `exclude` parameters for the time being - this is a breaking change and I'd rather have a bit more wiggle room to update other systems that may be using this.

Please add a comment that these filter properties are deprecated!

sergei-maertens (approved):

PR #374, opened 2021-07-16
Retrieve documents from dowc

damm89:


backend/src/zac/contrib/dowc/api.py
```diff
@@ -73,6 +75,19 @@ def create_doc(
...
+@optional_service
+def get_open_documenten(user: User, referer: str) -> List[Optional[DowcResponse]]:
+    client = get_client(user)
+    operation_id = "documenten_list"
+    kwargs = {
+        "purpose": DocFileTypes.write,
+        "info_url": referer,
+    }
+    url = get_operation_url(client.schema, operation_id, **kwargs)
+    response = client.request(url, operation_id, method="GET", expected_status=200)
```
sergei-maertens, line 87: Is this right? Afaik the kwargs are used for interpolation in the URL template, e.g. `/foo/{id}`. I thought the filter parameters worked as querystring parameters?

sergei-maertens (changes requested):

backend/src/zac/contrib/dowc/api.py
```diff
@@ -73,6 +74,26 @@ def create_doc(
...
+        {
+            "purpose": DocFileTypes.write,
+            "info_url": referer,
+        }
+    )
+    try:
+        response = client.request(
+            url.url, operation_id, method="GET", expected_status=200
+        )
+        return factory(DowcResponse, response)
```
sergei-maertens, line 92: you can keep this as is - but the `gemma-zds-client` exposes the requests interface (see https://github.com/VNG-Realisatie/gemma-zds-client/blob/master/zds_client/client.py#L181) so the cleanest approach would probably be:

```suggestion
    query = {
        "purpose": DocFileTypes.write,
        "info_url": referer,
    }
    try:
        response = client.request(
            url, operation_id, method="GET", expected_status=200, params=query
        )
        return factory(DowcResponse, response)
```

sergei-maertens (approved):

PR #376, opened 2021-07-16
Feature/hijack header

annashamray:
Issue - https://github.com/GemeenteUtrecht/ZGW/issues/835 (BE)

### DIFF
--- a/backend/src/zac/accounts/middleware.py
+++ b/backend/src/zac/accounts/middleware.py
@@ -12,6 +12,6 @@ class HijackMiddleware:
 
         # add hijack header for FE
         is_hijacked = request.session.get("is_hijacked_user", False)
-        response[self.header] = is_hijacked
+        response[self.header] = "true" if is_hijacked else "false"
 
         return response
--- a/backend/src/zac/accounts/middleware.py
+++ b/backend/src/zac/accounts/middleware.py
@@ -1,0 +1,17 @@
+from django.conf import settings
+
+
+class HijackMiddleware:
+    header = settings.HIJACK_HEADER
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        # add hijack header for FE
+        is_hijacked = request.session.get("is_hijacked_user", False)
+        response[self.header] = is_hijacked
+
+        return response
--- a/backend/src/zac/accounts/tests/test_hijack_header.py
+++ b/backend/src/zac/accounts/tests/test_hijack_header.py
@@ -1,0 +1,54 @@
+from django.conf import settings
+from django.test import TestCase, override_settings
+from django.urls import reverse, reverse_lazy
+
+from rest_framework import status
+
+from zac.elasticsearch.tests.utils import ESMixin
+
+from ..models import User
+from .factories import StaffUserFactory
+
+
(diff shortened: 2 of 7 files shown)


backend/src/zac/accounts/middleware.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        # add hijack header for FE
+        is_hijacked = request.session.get("is_hijacked_user", False)
+        response[self.header] = is_hijacked
```
sergei-maertens, line 15:
```suggestion
        response[self.header] = "true" if is_hijacked else "false"
```

Let's not rely on implicit type conversions, this is also consistent with https://github.com/GemeenteUtrecht/zaakafhandelcomponent/blob/main/backend/src/zac/contrib/kownsl/views.py#L110

backend/src/zac/api/schema.py
```diff
@@ -35,3 +39,16 @@ def get_summary(self):
...
         return action_or_method_summary or cls_summary
+
+    def get_override_parameters(self):
+        """ Add hijack header"""
+        hijack_response_header = OpenApiParameter(
+            name=settings.HIJACK_HEADER,
+            type=str,
+            location=OpenApiParameter.HEADER,
+            description=_("Header displays if the user is hijacked."),
+            enum=["False", "True"],
```
sergei-maertens, line 50:
```suggestion
            enum=["false", "true"],
```

sergei-maertens (changes requested):

backend/src/zac/accounts/tests/test_hijack_header.py
```diff
@@ -0,0 +1,54 @@
...
+
+        for url in self.urls:
+            with self.subTest(url):
+                response = self.client.get(url)
+
+                self.assertEqual(
+                    response.status_code, status.HTTP_200_OK, response.json()
+                )
+                self.assertIn(settings.HIJACK_HEADER, response)
+                self.assertEqual(response[settings.HIJACK_HEADER], "True")
```
damm89, line 54: After @sergei-maertens' suggestions: 
```suggestion
                self.assertEqual(response[settings.HIJACK_HEADER], "true")
```

backend/src/zac/accounts/tests/test_hijack_header.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def test_response_not_hijacked(self):
+        for url in self.urls:
+            with self.subTest(url):
+
+                response = self.client.get(url)
+
+                self.assertEqual(response.status_code, status.HTTP_200_OK)
+                self.assertIn(settings.HIJACK_HEADER, response)
+                self.assertEqual(response[settings.HIJACK_HEADER], "False")
```
damm89, line 38: After @sergei-maertens' suggestions: 
```suggestion
                self.assertEqual(response[settings.HIJACK_HEADER], "false")
```

damm89 (approved):

sergei-maertens (changes requested):
~~I don't see any changes to the `zac.conf.base` module to add the `HIJACK_HEADER` setting and it doesn't look like a default setting from Hijack either. Can you add it?~~

I need more coffee before I review code

sergei-maertens (approved):

PR #377, opened 2021-07-16
Feature #366 - Relate objects to zaak (backend)

### LINKED ISSUE #366: Add relations to objects (backend)
These are the endpoints needed to be able to relate a zaak to an object in the Object API:

- [x] Endpoint to retrieve objecttypes from the Objecttype API. The objecttypes returned have multiple versions.
- [x] Endpoint that given an objecttype version retrieves the JSONSchema of that objecttype from the objecttype API. This contains the data attributes that are found in objects of that objecttype.
- [x] Endpoint that takes the objecttype, date, registration date, geometry or data attributes to retrieve objects from the Object API. The data attribute has to be formatted using the key__operator__value format.
- [x] Endpoint that takes an object, objecttype and a zaak and creates a ZaakObject in OpenZaak.
- [ ] Work on the API documentation

SilviaAmAm:
Fixes #366


backend/src/zac/core/api/serializers.py
```diff
@@ -638,3 +638,59 @@ class UserAtomicPermissionSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
         fields = ("username", "permissions")
+
+
+class ObjecttypeSerializer(serializers.Serializer):
```
sergei-maertens, line 643: You can let most of the serializers auto-inspect using https://github.com/maykinmedia/zgw-consumers/blob/master/zgw_consumers/drf/serializers.py#L20

It's very similar to DRFs built-in `ModelSerializer`

backend/src/zac/core/api/views.py
```diff
@@ -678,3 +691,129 @@ def get_eigenschappen(self, zaaktypen):
...
+    # },
+)
+class ObjecttypeListView(ListAPIView):
+    authentication_classes = (authentication.SessionAuthentication,)
+    permission_classes = (permissions.IsAuthenticated,)
+    serializer_class = ObjecttypeSerializer
+    action = "list"
+
+    def get_queryset(self):
+        return fetch_objecttypes()
```
sergei-maertens, line 719: Please do not shoe-horn non-model things into Generic views/viewsets that are meant to work with models. It's confusing, since the return type is not as expected (list vs. queryset) and drf-spectacular is confused by it as well.

Better to just implement a pattern like in Open Forms: https://github.com/open-formulieren/open-forms/blob/master/src/openforms/utils/api/views.py

sergei-maertens (changes requested):

backend/src/zac/core/api/views.py
```diff
@@ -679,3 +693,74 @@ def get_eigenschappen(self, zaaktypen):
...
+
+#
+# Objects endpoints
+#
+
+
+@extend_schema(
+    summary=_("List Objecttypes"),
+    description=_("Retrieves all object types from the configured Objecttypes API."),
+)
```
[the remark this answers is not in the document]
SilviaAmAm: I added a tag `objects` since some of the enpoints are in the Objecttype API and one in the Zaken API

PR #381, opened 2021-07-19
Issues/2021 07 19

svenvandescheur:
A little bit more dry and maintainable. Something to think about: Could we generate this and the corresponding models of the API scheme?


sergei-maertens:
Generating the API models from the API docs is definitely possible, there exist generators. You'd have to check which angular/JS tooling there is for that, using OpenAPI 3 api docs.

kelvincy (approved):
looks good

PR #386, opened 2021-07-22
update atomic permission

### LINKED ISSUE #384: Update atomic permissions handlers
1. When the user is assigned as adviser or approver they are granted permissions to read the zaak, but they don't have permission to perform advice/approve task. 
2. When the user is assigned to the activity, they are granted permissions to read and update activities. but they don't have permission to read the zaak itself.

The atomic permission handlers and the documentation should be updated

annashamray:
fixes #384

### DIFF
--- a/backend/doc/authorization.rst
+++ b/backend/doc/authorization.rst
@@ -21,7 +21,7 @@ Each permission provides the right to perform one of the following types of oper
 
 * zaak permissions:
     * ``zaken:inzien`` - to see the case details. It's the most used permission in the ZAC
-    * ``zaken:wijzigen`` - to modify the case, for example, to change the confidential level
+    * ``zaken:wijzigen`` - to modify the case, for example, to change the confidentiality level
     * ``zaakproces:usertasks-uitvoeren`` - to perform Camunda tasks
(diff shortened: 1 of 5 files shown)


backend/doc/authorization.rst
```diff
@@ -21,7 +21,7 @@ Each permission provides the right to perform one of the following types of oper
 
 * zaak permissions:
     * ``zaken:inzien`` - to see the case details. It's the most used permission in the ZAC
-    * ``zaken:wijzigen`` - to modify the case, for example, to add documents to the case
+    * ``zaken:wijzigen`` - to modify the case, for example, to change the confidential level
```
damm89, line 24:
```suggestion
    * ``zaken:wijzigen`` - to modify the case, for example, to change the confidentiality level
```

damm89: Some more "mierenn*kerij":

Lines 45 and 58 end in periods when none of the other ones do. 
The bullet point list starting at line 155 has sentences that end in periods but don't start with an upper case.

damm89 (changes requested):
Please correct some of the formatting - otherwise looks good to me.

sergei-maertens (approved):

PR #395, opened 2021-08-03
Add django scim2

SilviaAmAm:
Added django-scim2 and expose `AuthorizationProfile` through the SCIM API.


backend/src/zac/accounts/migrations/0003_user_uuid.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ("accounts", "0002_auto_20200406_0932_squashed_0047_auto_20210709_1335"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="uuid",
+            field=models.UUIDField(default=uuid.uuid4, editable=False, unique=True),
+        ),
```
sergei-maertens, line 18: this is a migration that will fail with existing data, because the `default` is only evaluated once and then used for every existing record.

So, all existing users will get the same UUID which violates the `unique` constraint.

backend/src/zac/accounts/scim/adapters.py
```diff
@@ -0,0 +1,111 @@
...
+
+class UserAdapter(SCIMUser):
+    id_field = "uuid"
+
+    @property
+    def location(self):
+        return build_absolute_url(path=self.path, request=self.request)
+
+    def delete(self):
+        User.objects.filter(uuid=self.id).delete()
```
sergei-maertens, line 71: I'd prefer the user to be set to `is_active=False`, since a lot of data will probably be deleted in a cascade otherwise, and other systems may break since the username is referenced in places which will no longer exist so "FK" lookups will break.

Perhaps even introduce an explicit `deleted` flag on the user object, but keep the record in the database + retain existing data. We'll have to check with the GDPR people to figure out which data of the user should be wiped, but we don't keep sensitive data anyway - I don't see much issues with keeping the first and last name and (professional) e-mail address. I think audit trail purposes are far more important here than actually deleting the user.

backend/src/zac/conf/includes/base.py
```diff
@@ -518,3 +520,14 @@
     "default": {"hosts": config("ES_HOST", "localhost:9200")},
 }
 ES_INDEX_ZAKEN = "zaken"
+
+
+# SCIM
+SCIM_SERVICE_PROVIDER = {
+    "NETLOC": "localhost",
```
sergei-maertens, line 527: is this correct? What does it do/drive?

backend/src/zac/conf/includes/base.py
```diff
@@ -518,3 +520,14 @@
...
+    "AUTHENTICATION_SCHEMES": [],
```
sergei-maertens, line 528: It's unclear to me how authentication works. I can see from the test that auth is required, but it looks like it relies on `SessionAuthentication` now?

It would be cool if `TokenAuthentication` was possible.

sergei-maertens (changes requested):
Looks thorough - I have some remarks and questions though, can you address those?

backend/src/zac/accounts/migrations/0003_user_uuid.py
SilviaAmAm: Ah!! Now I changed the migrations following these steps: https://docs.djangoproject.com/en/3.2/howto/writing-migrations/#migrations-that-add-unique-fields

backend/src/zac/accounts/scim/adapters.py
SilviaAmAm: I ended up changing the `is_active` attribute. 
Now the a user with `is_active=False` will still appear when doing a GET `/scim/v2/Users` (but with an attribute `active=false`. It will also appear among the members of a group (displaying only the uuid, the url and the display name of the user). 
Should the latter maybe change to filter out inactive users?

backend/src/zac/conf/includes/base.py
SilviaAmAm: It is used to derive absolute URLs https://github.com/15five/django-scim2/blob/master/src/django_scim/utils.py#L103 so that the URLs of the different resources can be put in responses (https://github.com/15five/django-scim2/blob/master/src/django_scim/adapters.py#L69, https://github.com/15five/django-scim2/blob/master/src/django_scim/views.py#L263) 

I changed the `location` attribute of our adapters to build the URL with `build_absolute_url`, but since it is still used in the search views for the `Location`, I guess it should be changed to the domain name (or the search view should be overwritten to also build the URLs from the requests?)

backend/src/zac/conf/includes/base.py
SilviaAmAm: The `django_scim2` middleware checks if the `request` has a user attribute (and that the user is not anonymous). In this case it lets the request go through, otherwise not. It says in the documentation that this app does not implement authorisation/authentication and that this is left to other apps. They make the example of Django OAuth Toolkit, with:
```
'AUTHENTICATION_SCHEMES': [
        {
            'type': 'oauth2',
            'name': 'OAuth 2',
            'description': 'Oauth 2 implemented with bearer token',
        },
    ],
```

sergei-maertens: alright, we'll have to check that, I'd like to use DRFs TokenAuth for this.

sergei-maertens (approved):
I'll check the auth set-up and amend it to this PR, but otherwise this looks good.

backend/src/zac/conf/includes/base.py
sergei-maertens: I'm letting it default now to `ALLOWED_HOSTS[0]` if present, and you can also specify an envvar for this.

Probably if we do anything else with this, we'll rather change the default location getter to actually use the request object than override more adapters. But for now, this will do.

PR #400, opened 2021-08-11
:sparkles: Assign groups to user tasks

damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/836


backend/src/zac/camunda/api/views.py
```diff
@@ -252,22 +253,44 @@ class SetTaskAssigneeView(APIView):
...
+                    [part[0].upper() + "." for part in user.first_name.split()]
+                ).strip(),
+            }
+        else:
+            group = Group.objects.get(name=_name)
+            return {
+                "identificatie": group.name,
+                "achternaam": group.name,
+                "voorletters": "Groep",
+            }
```
sergei-maertens, line 273: This doesn't sit well, I think we should just skip this until someone from the group actually self-assigns the task.

backend/src/zac/camunda/data.py
```diff
@@ -45,3 +48,13 @@ def has_form(self) -> bool:
 
     def get_variable(self, name: str) -> Any:
         return get_task_variable(self.id, name)
+
+    def assignee_type(self) -> str:
+        if self.assignee:
+            from ..core.camunda.utils import User
```
sergei-maertens, line 54: that looks like a weird import, shouldn't that just be `zac.accounts.models.User`? And then probably you can lift the import to the top of the module.

backend/src/zac/core/camunda/utils.py
```diff
@@ -21,12 +22,20 @@
...
 
 
-def _resolve_assignee(username: str) -> User:
-    try:
-        user = User.objects.get(username=username)
-    except User.DoesNotExist:
-        user = User.objects.create_user(username=username)
-    return user
+def _resolve_assignee(name: str) -> Union[Group, User]:
+    user_or_group, _name = name.split(":")[0], ":".join(name.split(":")[1:])
```
sergei-maertens, line 26: Can be simplified a bit, no?

```py
>>> "foo:bar:baz".split(":", 1)
['foo', 'bar:baz']
```

```suggestion
    user_or_group, _name = name.split(":", 1)
```

backend/src/zac/core/camunda/utils.py
```diff
@@ -21,12 +22,20 @@
...
+    if user_or_group == "group":
```
sergei-maertens, line 27: can we re-use the `AssigneeTypes` choices/constants instead?

backend/src/zac/core/camunda/utils.py
```diff
@@ -21,12 +22,20 @@
...
+        try:
+            group = Group.objects.get(name=_name)
+        except Group.DoesNotExist:
+            group = Group.objects.create(name=_name)
```
sergei-maertens, line 31: Not a bad idea tbh, maybe add a `logger.info(...)` statement so we can check logs for why a group was created if we ever wonder about that?

backend/src/zac/core/services.py
```diff
@@ -692,15 +693,26 @@ def update_medewerker_identificatie_rol(zaak: Zaak) -> Optional[List[Rol]]:
...
+                        factory(Rol, new_rol)
+                    ) != get_naam_medewerker(rol):
+                        new_rollen.append((rol.url, new_rol))
+                except User.DoesNotExist:
+                    group = Group.objects.get(name=identificatie)
+                    continue
+                    # Do nothing if "medewerker" is a group.
+                    # Groups are only currently assigned roles within the
+                    # zac with all the information needed to proceed.
+                    # TODO: Create group roles?
```
sergei-maertens, line 711: I'd even go so far as to delete the `Rol` if it's a group, it makes no sense in a ZGW context.

There's a construction with "Organizational unit" that *may* be relevant, but we should check that with information architects and the rest of the team.

sergei-maertens (changes requested):
I think it's a nasty feature in general and we have to accept the nature of this "hack".

The polymorphic serializer is a nice approach to getting some structure in the system, good idea!

backend/src/zac/camunda/data.py
damm89: Yep.

backend/src/zac/core/services.py
```diff
@@ -3,6 +3,7 @@
 from typing import Any, Dict, List, Optional, Tuple, Union
 from urllib.parse import urljoin
 
+from django.contrib.auth.models import Group
 from django.core.cache import cache
```
damm89, line 7:
```suggestion
from django.core.cache import cache
```

sergei-maertens (approved):

PR #407, opened 2021-08-17
Feature/role api

annashamray:
issue - GemeenteUtrecht/ZGW#773
Depends on #403 

**Done in the PR:**
* add GET, POST and PUT endpoints for `Role`
* GET endpoint to list permissions

### DIFF
--- a/backend/src/zac/accounts/api/views.py
+++ b/backend/src/zac/accounts/api/views.py
@@ -1,11 +1,16 @@
 from django.http import JsonResponse
+from django.utils.translation import ugettext_lazy as _
 
+from drf_spectacular.utils import extend_schema
 from rest_framework import views
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.permissions import IsAuthenticated
+from rest_framework.response import Response
 
-from zac.accounts.api.serializers import CatalogusURLSerializer
 from zac.core.services import get_informatieobjecttypen
+
+from ..permissions import registry
+from .serializers import CatalogusURLSerializer, PermissionSerializer
 
 
 class InformatieobjecttypenJSONView(views.APIView):
@@ -37,3 +42,18 @@ class InformatieobjecttypenJSONView(views.APIView):
                 }
             )
         return JsonResponse(response_data)
+
+
+class PermissionView(views.APIView):
+    authentication_classes = [SessionAuthentication]
+    permission_classes = [IsAuthenticated]
+    serializer_class = PermissionSerializer
+
+    @extend_schema(
+        summary=_("List permissions"), responses={200: serializer_class(many=True)}
+    )
+    def get(self, request):
+        """Return all available permissions and their description"""
+        permissions = list(registry.values())
+        serializer = self.serializer_class(permissions, many=True)
+        return Response(serializer.data)
--- a/backend/src/zac/accounts/api/views.py
+++ b/backend/src/zac/accounts/api/views.py
@@ -5,8 +5,8 @@ from drf_spectacular.utils import extend_schema
 from rest_framework import views
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.permissions import IsAuthenticated
-from rest_framework.response import Response
 
+from zac.core.api.mixins import ListMixin
 from zac.core.services import get_informatieobjecttypen
 
 from ..permissions import registry
@@ -44,16 +44,14 @@ class InformatieobjecttypenJSONView(views.APIView):
         return JsonResponse(response_data)
 
 
-class PermissionView(views.APIView):
+@extend_schema(
+    summary=_("List permissions"),
+    description=_("Return all available permissions and their description"),
+)
+class PermissionView(ListMixin, views.APIView):
     authentication_classes = [SessionAuthentication]
     permission_classes = [IsAuthenticated]
     serializer_class = PermissionSerializer
 
-    @extend_schema(
-        summary=_("List permissions"), responses={200: serializer_class(many=True)}
-    )
-    def get(self, request):
-        """Return all available permissions and their description"""
-        permissions = list(registry.values())
-        serializer = self.serializer_class(permissions, many=True)
-        return Response(serializer.data)
+    def get_objects(self):
+        return list(registry.values())
--- a/backend/bin/generate_schema.sh
+++ b/backend/bin/generate_schema.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+src/manage.py spectacular --file src/openapi.yaml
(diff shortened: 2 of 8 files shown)


damm89 (approved):
:+1:

av314:
In last Fridays dev call I raised my reservation about the use of enums for status types in the API (not sure if it was really ‘status types’, but that doesn’t really matter for this discussion).

The set of status types in the API, is not fixed, but may be extended in the future.
For multiple reasons using a string for these status types is preferred over enums.
The API design already provides a GET call that returns all allowed/defined status types.
Using the result of this GET call to validate a status type is a much better approach than relying on the set of enum values.
This means that the API does not have to be changed when a new status type is added.

Common Ground advocates an architecture of loosely-coupled components, whereby components can be developed and deployed independent of each other.
Adding or changing a status value should be gracefully handled if client and server (temporarily) use different API versions. This is the case if the status type is a string, but not necessarily for enums - which are per definition closed sets of values.

A typical application/component consists of three layers (1) API/controller, (2) Service/business logic and (3)  Persistance.
The API/controller should do ‘superficial’ validations like: are all mandatory fields present, is the email field in the right format. Checking whether the value of a field is valid belongs in the Service/business logic layer.

If we use the GET call on the propagating the list of status types to the client also has the advantage that the status types are only defined in one place and not spread throughout the code. Having a ‘single source of truth’ means that there is no risk that you forget to update all references to an enum.

svenvandescheur:
> If we use the GET call on the propagating the list of status types to the client also has the advantage that the status types are only defined in one place and not spread throughout the code. Having a ‘single source of truth’ means that there is no risk that you forget to update all references to an enum.
 
I think in practice these kind of information is stored somewhere in a model and exposed using one ore more API's. The AP endpoint acts as a view on this information and is not a source on it's own. If we were to only expose the choices through a dedicated endpoint and use pure strings instead we might loose the off the shelf validation that Django/DRF gives us. Instead, we might end up validation the choices our selves with al risks of bugs as a result.

_* Let's have the backend team explain this._

This validation should then be done one two places, backend and frontend. And especially the later doesn't really benefit from these "optimisations" since we should now do two calls, one for the scheme and one for the data. In a way, you'd still have two sources of truth since the value in API endpoint one should match the choices in API endpoint two. A check that's may be a little bit more complicated since it relies on asynchronous behaviour.

I think the point here is that although theoretically perfectly understandable, in reality these kind of practices makes it very difficult to actual implement things in a frontend and IMO doesn't do what I think they're intended to do: make things better.

The BFF stands for backend for frontend, it's only purpose is to provide the frontend with information endpoints. I'd imagine that the nicest situation would be where this actually functions in such a way that its suits the frontend, it's sole customer. I think these kind of BFF justify a slightly different approach in architecture since the architecture used for services is very inefficient for building frontends.

Even through I don't like the following setups: many bff's work in such a way that they provide unschemed, unstructured context specific for a specific view/component. All the way to graphql where the frontend, instead of the backend dictates what information is returned (and how). The benefit of this approach is there is very little overhead and data processing in the frontend is limited to a minimum.

I personally rather work, much like we do know with endpoints for objects, but without losing the reality of having to work with the information (like generating options from an enum). The possibility of generating components from the response of a single API endpoint is a big thing here.

Maybe we should consider differentiating the guidelines for specification service API's and bffs which IMO are two entirely different things.

- service API: as proposed, the most semantically correct RESTful approach.
- backend for frontend: semantically correct RESTful approach with integrated surrounding context.

sergei-maertens:
> * Let's have the backend team share their knowledge on this.

You summarized it perfectly in my opinion. It costs us extra work and extra risks to strip the enum out of the API schema, since this is _generated_ from the restframework `serializers.ChoiceField` which has the validation baked in. And because it's a standard ChoiceField, drf-spectacular automatically extracts the values and generates an enum for it.

I think we should not pretend as if there will be a new permission code/string every week. The defined, existing permissions are fairly stable, I would even dare say that they are more stable than the schema where we are adding or tweaking endpoints to meet the frontend-requirements.

I also 100% agree with the idea that this is an internal/private API specification and the schema/documentation serves to make it easy for frontend-developers to know what there is that they can use and so that they can request missing functionality from backend. The ZAC is drawn on layers 4 and 5 of the Common Ground model, not on layer 2, where these concerns would be much more substantiated.

@av314 

> For multiple reasons using a string for these status types is preferred over enums.

I'm not sure if you are listing the reasons below that line, or if this is just a broad statement. If it's the latter, can you provide a verifiable list of reasons?

annashamray:
To be honest I don't really understand what is the source of disagreement: only the API spec or the API spec + API logic? 

Do we all agree with the logic, that the role endpoint validates if the permission is of the list of available choices and if it's not true, 400 error will appear? 

If not, than what behavior is expected from the backend if the permission string is not correct? 

If yes, then the whole discussion is about the API document. The main client of this document is frontend developers, so let's do it the way they find it the easiest to understand and work with. And we keep in mind that permission choices are updated automatically and simultaneously in the `/permissions` endpoint and in the API spec.

backend/src/zac/accounts/api/views.py
```diff
@@ -37,3 +42,18 @@ def get(self, request):
                 }
             )
         return JsonResponse(response_data)
+
+
+class PermissionView(views.APIView):
```
sergei-maertens, line 47: You can use `zac.core.api.mixins.ListMixin` to make this more declarative.

(yes, the mixin may be moved to `zac.api` instead too)

If yo do that, you no longer need to specify `responses` in the `@extend_schema` call

sergei-maertens (changes requested):
Small change related to API schema generation

backend/src/zac/accounts/api/views.py
annashamray: Thank you! I didn't know about this mixin

av314:
@sergei-maertens, you wrote: "_I'm not sure if you are listing the reasons below that line, or if this is just a broad statement. If it's the latter, can you provide a verifiable list of reasons?_"

I am listing four arguments further on in my initial comment. 

My comment is about best practices for API design, these are not things that I made up myself.
Best practices not only aply to Layer 2, but to any API.

av314:
@annashamray , the API is not WRONG or so, I am just giving feedback for improvement.
Making the API more relilient, fault tollerant etc ...
The fact thay you have a /permissions endpoint is good, but this endpoint is not really used by the client.

sergei-maertens (approved):

annashamray:
@av314 

> The fact thay you have a /permissions endpoint is good, but this endpoint is not really used by the client.

Do you have this information from our FE developers? If this endpoint is not used by FE I'd rather remove it. I don't want to produce some code which is not used. 
@svenvandescheur Could you please take a look to consider if FE is going to use `/permissions` endpoint

av314:
@annashamray my assumption that the /permisions endpoint is currently not used is based on discussions.
I suggest to keep this endpint for future use.

annashamray:
As discussed in the standup I merge the PR

PR #409, opened 2021-08-19
Feature/auth profile users api

annashamray:
depends on #403


annashamray:
This change wasn't requested, close PR for now

Issue #413, opened 2021-08-20
Decompose ES search


annashamray:
ES search is keeping to expand and soon it would be hard to maintain it, 
I think we can decompose\refactor it

annashamray:
@damm89 @sergei-maertens Shall we discuss it?

PR #416, opened 2021-08-24
Increase index_zaken code coherence and performance

damm89:
Related to issue: #413.
I think the creation of elasticsearch documents should be separated from the saving and indexing of those documents. 

This is relevant for the bulk API of elasticsearch as well as making the code a little bit easier to handle (although more verbose). The creation and appending of inner docs is more explicit and the purpose of the related functions a bit more clear and aligned.

Performance changes (on local machine) based on ~6300 zaken with related objects.

- Baseline:
> - Time: ~3000 seconds
> - Peak memory usage of python process: baseline_index
- Bulk (without batch):
> - Time: ~700 seconds
> - Peak memory usage of python process: bulk_index
- Bulk (with batch):
> - Time: ~1000 seconds
> - Peak memory usage of python process:bulk_batch

### DIFF
--- a/backend/src/zac/elasticsearch/api.py
+++ b/backend/src/zac/elasticsearch/api.py
@@ -10,6 +10,7 @@ from zgw_consumers.api_models.zaken import Status, ZaakEigenschap, ZaakObject
 from zac.core.rollen import Rol
 from zac.core.services import (
     get_rollen,
+    get_status,
     get_statustype,
     get_zaak_eigenschappen,
     get_zaakobjecten,
@@ -63,6 +64,8 @@ def _get_zaak_document(
             return
         else:
             zaak_document = create_zaak_document(create_zaak)
+            zaak_document.save()
+
     return zaak_document
 
 
@@ -129,11 +132,12 @@ def create_status_document(status: Status) -> StatusDocument:
 
 
 def update_status_in_zaak_document(zaak: Zaak) -> None:
-    status_document = create_status_document(zaak.status) if zaak.status else None
-
-    zaak_document = _get_zaak_document(zaak.uuid, zaak.url, create_zaak=zaak)
-    zaak_document.status = status_document
-    zaak_document.save()
+    if zaak.status:
+        zaak.status = get_status(zaak) if isinstance(zaak.status, str) else zaak.status
+        status_document = create_status_document(zaak.status)
+        zaak_document = _get_zaak_document(zaak.uuid, zaak.url, create_zaak=zaak)
+        zaak_document.status = status_document
+        zaak_document.save()
 
     return
 
--- a/backend/src/zac/elasticsearch/api.py
+++ b/backend/src/zac/elasticsearch/api.py
@@ -183,23 +183,19 @@ def update_eigenschappen_in_zaak_document(zaak: Zaak) -> None:
     return
 
 
-def create_zaakobjecten_document(
-    zaakobjecten: List[ZaakObject],
-) -> List[ZaakObjectDocument]:
-    return [
-        ZaakObjectDocument(
-            url=zo.url,
-            object=zo.object,
-        )
-        for zo in zaakobjecten
-    ]
+def create_zaakobject_document(
+    zaakobject: ZaakObject,
+) -> ZaakObjectDocument:
+    return ZaakObjectDocument(url=zaakobject.url, object=zaakobject.object)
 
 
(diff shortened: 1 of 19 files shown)


backend/src/zac/elasticsearch/api.py
```diff
@@ -126,94 +97,104 @@ def delete_zaak_document(zaak_url: str) -> None:
...
 
     return
 
 
-def update_zaakobjecten_in_zaak_document(zaak: Zaak) -> None:
-    zaak_document = _get_zaak_document(zaak.uuid, zaak.url, create_zaak=zaak)
-    zaak_document.objecten = [
+def create_zaakobjecten_document(
+    zaakobjecten: List[ZaakObject],
+) -> List[ZaakObjectDocument]:
```
annashamray, line 183: Some nitpicking: `create_zaakobjecten_document` and `create_rol_document` are not quite consistent with each other, since the first function creates the list of documents, and the latter - only one document. 

It would be nice if they have the same structure, cause `rollen` and `zaakobjecten` documents have the similar shape

backend/src/zac/elasticsearch/management/commands/index_zaken.py
```diff
@@ -20,40 +44,111 @@ def handle(self, **options):
...
+        # Build the zaak_documenten
+        zaak_documenten = {zaak.url: create_zaak_document(zaak) for zaak in zaken}
+        return zaak_documenten
+
+    def create_zaaktype_documenten(
+        self, zaken: List[Zaak]
+    ) -> Dict[str, ZaakTypeDocument]:
+        unfetched_zaaktypen = [
+            zaak.zaaktype for zaak in zaken if isinstance(zaak.zaaktype, str)
+        ]
```
annashamray, line 85:
```suggestion
        unfetched_zaaktypen = {
            zaak.zaaktype for zaak in zaken if isinstance(zaak.zaaktype, str)
        }
```

annashamray (commented):
I really like the pattern created in `elasticsearch/api.py` where you separate creating ES documents from fetching data from external APIs. It really helps with reusing components and reminds about single-responsibility principle :) 

I'll approve it when CI turns green again

backend/src/zac/elasticsearch/api.py
damm89: yes, good one

backend/src/zac/elasticsearch/management/commands/index_zaken.py
```diff
@@ -1,59 +1,180 @@
...
+                zaken, query_params = get_zaken_all_paginated(
+                    client, query_params=query_params
+                )
+                get_more = query_params.get("page", None)
+                for zaak in zaken:
+                    zaak.zaaktype = zaaktypen[zaak.zaaktype]
+
+                yield from self.zaakdocumenten_generator(zaken)
+
+    def zaakdocumenten_generator(self, zaken: List[Zaak]) -> List[ZaakDocument]:
```
sergei-maertens, line 64:
```suggestion
    def zaakdocumenten_generator(self, zaken: List[Zaak]) -> Iterator[ZaakDocument]:
```

backend/src/zac/elasticsearch/management/commands/index_zaken.py
```diff
@@ -1,59 +1,180 @@
...
-        self.index_zaak_eigenschappen(zaken)
+    def handle(self, **options):
+        self.clear_zaken_index()
+        bulk(
+            connections.get_connection(),
+            self.batch_index_zaken(),
+        )
 
-        self.stdout.write("Zaken have been indexed")
+        count = ZaakDocument.search().count()
```
sergei-maertens, line 89: isn't this limited to 1000 by default? Unless it optimizes it under the hood to do the equivalent of a `SELECT COUNT(*)`

sergei-maertens (changes requested):

backend/src/zac/elasticsearch/management/commands/index_zaken.py
sergei-maertens: I don't know _how_, but it seems fine:

```
>>> from zac.elasticsearch.searches import *
>>> ZaakDocument
<class 'zac.elasticsearch.documents.ZaakDocument'>
>>> ZaakDocument.search().count()
2500
```

sergei-maertens (approved):

PR #433, opened 2021-09-07
add DELETE endpoint for zaak objects

annashamray:
Issue - https://github.com/GemeenteUtrecht/ZGW/issues/900

I don't think that PATCH or PUT endpoint is needed here. Despite we now have PATCH support in Zaken API, it doesn't include object field, which can't be changed. So we can't change the object in the zaakobject. To do so you indeed need to delete the old object and to create the new one. It's the only way which is supported by Zaken API.


sergei-maertens (approved):

damm89 (approved):
After clarification in slack :-)

annashamray:
Part of: #442

PR #439, opened 2021-09-14
:bug: Fix empty form_key in user task bug

damm89:
Resolves https://github.com/GemeenteUtrecht/ZGW/issues/874

### DIFF
--- a/backend/src/zac/camunda/api/serializers.py
+++ b/backend/src/zac/camunda/api/serializers.py
@@ -4,10 +4,11 @@ from django.contrib.auth.models import Group
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
+from rest_framework.fields import _UnvalidatedField
 
 from zac.accounts.api.serializers import GroupSerializer, UserSerializer
 from zac.accounts.models import User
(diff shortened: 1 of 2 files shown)


backend/src/zac/camunda/api/serializers.py
```diff
@@ -67,9 +68,8 @@ class ProcessInstanceSerializer(serializers.Serializer):
     tasks = TaskSerializer(many=True)
 
 
-class ChoiceFieldNoValidation(serializers.ChoiceField):
-    def to_internal_value(self, data):
-        return data
+class ChoiceFieldNoValidation(_UnvalidatedField, serializers.ChoiceField):
```
annashamray, line 71: Tbh I'm not a huge fan of using protected class which is used only for composite fields in `rest_framework`. 
Could `allow_null` attribute fix the issue?

damm89: Fair enough! allow_null won't suffice :-) Removed the protected class.

sergei-maertens (approved):

annashamray (approved):
looks good!

PR #441, opened 2021-09-16
Feature/add zaakobject url

annashamray:


svenvandescheur:
Part of: https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/442

annashamray:
Part of: #442

backend/src/zac/core/zaakobjecten.py
```diff
@@ -36,17 +37,23 @@ def retrieve_items(self, items: Iterator[ZaakObject]) -> None:
...
+        objects_in_object_api = fetch_objects(list(object_url_in_object_api))
+        object_url_mapping = {obj["url"]: obj for obj in objects_in_object_api}
 
-        for item in self.items:
-            if isinstance(item, str) and item.startswith(object_api.api_root):
-                objects_in_object_api.append(item)
-            else:
-                other_objects.append(other_objects)
+        self.items = []
+        for zaakobject_url, item in object_items:
```
damm89, line 49: This will raise an exception if item on line 31 is not a tuple (which it shouldn't be if the typing hint is correct).

damm89 (changes requested):

PR #443, opened 2021-09-20
:sparkles: Add max_workers argument to index_zaken command

damm89:
This seems to be the most straightforward method of keeping memory use stable and in check while still having better performance than original (~200% faster but also >200% slower than keeping max_workers at 10). 

I've scripted the indexing to be run with a range of max workers so you can see changes to the real and virtual memory and the total time until completion in each case.

You can easily see that there is no rampant increase in memory use throughout the entire indexing. Virtual memory increases the most. Using docker-compose and the following docker-compose.yml I'm actually unable to reproduce any OOM errors at all even though virtual memory is 0 in these cases.

```
version: '3.4'

services:
  db:
    image: postgres:11
    environment:
      - POSTGRES_USER=${DB_USER:-postgres}
      - POSTGRES_PASSWORD=${DB_PASSWORD:-postgres}
    volumes:
      - postgres_data:/var/lib/postgresql/data

  redis:
    image: redis:5-alpine

  elasticsearch:
    image: docker.elastic.co/elasticsearch/elasticsearch:7.9.2
    environment:
      - discovery.type=single-node
      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
    volumes:
      - es_data:/usr/share/elasticsearch/data
    # expose to the host machine so that we can use docker ES for development
    ports:
      - 9200:9200
      - 9300:9300

  web:
    image: scrumteamzgw/zaakafhandelcomponent:${TAG:-latest}
    build: .
    environment:
      - AUTORELOAD=true
      - DJANGO_SETTINGS_MODULE=zac.conf.docker
      - SECRET_KEY=${SECRET_KEY:-changeme}
      - IS_HTTPS=0
      - ALLOWED_HOSTS=localhost
      - REDIS_HOST=redis
      - ES_HOST=elasticsearch
      - USE_REDIS_CACHE=True
      - CACHE_DEFAULT=redis:6379/0
      - CACHE_AXES=redis:6379/0
      - CACHE_OAS=redis:6379/1
      - CACHE_SESSIONS=redis:6379/1
      - CORS_HEADERS_ENABLED=True
      - DEBUG=True
    # expose backend port for frontend development
    ports:
      - 8000:8000
    depends_on:
      - db
      - redis
      - elasticsearch
    memswap_limit: 1G
    mem_limit: 1G
    mem_reservation: 1G

volumes:
  postgres_data:
  es_data:
```
**Max_workers = 50**
index_zaken_max_workers_50
**Max_workers = 25**
index_zaken_max_workers_25
**Max_workers = 10**
index_zaken_max_workers_10
**Max_workers = 7**
index_zaken_max_workers_7
**Max_workers = 5**
index_zaken_max_workers_5
**Max_workers = 4**
index_zaken_max_workers_4

### DIFF
--- a/backend/src/zac/elasticsearch/management/commands/index_zaken.py
+++ b/backend/src/zac/elasticsearch/management/commands/index_zaken.py
@@ -2,6 +2,7 @@ from typing import Dict, Iterator, List
 
 from django.conf import settings
 from django.core.management import BaseCommand
+from django.core.management.base import CommandParser
 
 from elasticsearch.helpers import bulk
 from elasticsearch_dsl import Index
@@ -42,7 +43,16 @@ from ...documents import (
 class Command(BaseCommand):
     help = "Create documents in ES by indexing all zaken from ZAKEN API"
 
+    def add_arguments(self, parser: CommandParser) -> None:
+        super().add_arguments(parser)
+        parser.add_argument(
+            "--max_workers",
+            type=int,
+            help="Indicates the max number of parallel workers (for memory management). Defaults to 4.",
+        )
+
     def handle(self, **options):
+        self.max_workers = options.get("max_workers", 4)
         self.clear_zaken_index()
         bulk(
             connections.get_connection(),
@@ -104,7 +114,7 @@ class Command(BaseCommand):
         unfetched_zaaktypen = {
             zaak.zaaktype for zaak in zaken if isinstance(zaak.zaaktype, str)
         }
-        with parallel(max_workers=10) as executor:
+        with parallel(max_workers=self.max_workers) as executor:
             results = executor.map(fetch_zaaktype, unfetched_zaaktypen)
         zaaktypen = {zaaktype.url: zaaktype for zaaktype in list(results)}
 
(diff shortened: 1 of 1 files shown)


annashamray:
@damm89 Could you also add charts with memory usage as is and with this change please?

backend/src/zac/elasticsearch/management/commands/index_zaken.py
```diff
@@ -42,7 +43,16 @@
...
+    def add_arguments(self, parser: CommandParser) -> None:
+        super().add_arguments(parser)
+        parser.add_argument(
+            "--max_workers",
+            type=int,
+            help="Indicates the max number of parallel workers (for memory management). Defaults to 4.",
+        )
+
     def handle(self, **options):
+        self.max_workers = options.get("max_workers", 4)
```
sergei-maertens, line 55: you can move the default to the `add_argument` call itself

sergei-maertens (changes requested):

annashamray (approved):
Let's try decreasing workers number and see what happens

sergei-maertens (approved):

PR #444, opened 2021-09-22
:sparkles: Manage users in groups in django admin

damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/912

### DIFF
--- a/backend/src/zac/accounts/admin.py
+++ b/backend/src/zac/accounts/admin.py
@@ -1,6 +1,6 @@
 from django import forms
 from django.contrib import admin
-from django.contrib.auth.admin import UserAdmin
+from django.contrib.auth.admin import GroupAdmin as _GroupAdmin, UserAdmin
 from django.contrib.auth.models import Group
 from django.contrib.postgres.fields import JSONField
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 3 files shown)


backend/src/zac/accounts/admin.py
```diff
@@ -191,3 +193,15 @@ def formfield_for_dbfield(self, db_field, request, **kwargs):
...
+# Unregister old GroupAdmin
+admin.site.unregister(Group)
+
+# Register new GroupAdmin
+@admin.register(Group)
+class GroupAdmin(_GroupAdmin):
+    form = GroupAdminForm
+
+    def get_queryset(self, request):
+        return super().get_queryset(request).prefetch_related("user_set")
```
annashamray, line 207:
```suggestion
        return super().get_queryset(request).prefetch_related("user_set")
        
    def get_users_display(self, obj):
        return self.display_related_as_list_of_links(obj, "user_set")

    get_users_display.short_description = _("users")
```

There is only the name of the groups in the list display, I think it would be nice to also add the list of the users there. `RelatedLinksMixin` can be used for it.

annashamray (approved):
TIL that `filter_horizontal` doesn't work with reverse m2m relations. Nice workaround for it! 
I think it would be also nice to have inlines with groups in the `User` admin, but it's optional

backend/src/zac/accounts/admin.py
damm89: What if you get like 20 users in a group? Wouldn't that be super messy?

annashamray: Yeah you are right. It'd better the other way around but user's list page is already very busy. So please skip this coment

sergei-maertens (approved):

PR #449, opened 2021-09-29
Issue/fix rol notif handler

### LINKED ISSUE #448: Fix updating rols when notifications are received
Recently we experienced bursts of thousands of rol notifications that caused the collapse of OZ, ON and ZAC services. 
 
The root cause is the updating of "betrokkene" for a given zaak where we add the initials + last name of the "medewerker" where only an identification is referenced. The instances zac-test and zac-acc are competing with each other and triggering notifications for each other, which goes into an infinite loop.

The error and options how to solve it are described here - https://gist.github.com/sergei-maertens/888c2e744a24e96e31b1ba78e1d2f2ad 

**The 1st option is chosen:**
* ZAC will try to update notifications only when receiving `create` rol notifications
* ZAC will update only one rol. 
* If the rol already contains non-empty name attributes it is not updated
* If the related user in ZAC doesn't have name attributes. the rol is not updated
* The suggested hash is not implemented yet.

annashamray:
fix #448

### DIFF
--- a/backend/src/zac/core/services.py
+++ b/backend/src/zac/core/services.py
@@ -741,7 +741,7 @@ def update_medewerker_identificatie_rol(rol_url: str) -> Optional[Rol]:
     try:
         user = User.objects.get(username=identificatie)
     except ObjectDoesNotExist:
-        logger.warning("Couldn't find user with identificatie %s" % identificatie)
+        logger.warning("Couldn't find user with identificatie %s", identificatie)
         return
     if not user.get_full_name():
         return
--- a/backend/src/zac/core/services.py
+++ b/backend/src/zac/core/services.py
@@ -720,39 +720,41 @@ def update_rol(rol_url: str, new_rol: Dict) -> Rol:
     return factory(Rol, rol)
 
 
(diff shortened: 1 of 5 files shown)


backend/src/zac/core/services.py
```diff
@@ -720,39 +720,41 @@ def update_rol(rol_url: str, new_rol: Dict) -> Rol:
...
+        or rol.betrokkene_identificatie["achternaam"]
+    ):
+        return
+
+    # Try to get user data
+    identificatie = rol.betrokkene_identificatie["identificatie"]
+    try:
+        user = User.objects.get(username=identificatie)
+    except ObjectDoesNotExist:
+        logger.warning("Couldn't find user with identificatie %s" % identificatie)
```
sergei-maertens, line 744:
```suggestion
        logger.warning("Couldn't find user with identificatie %s", identificatie)
```

sergei-maertens (changes requested):
Small logger fix, other than that LGTM!

damm89 (approved):
:+1:

sergei-maertens (approved):

PR #454, opened 2021-10-08
Feature/group activities

damm89:


backend/src/zac/werkvoorraad/api/views.py
```diff
@@ -72,6 +78,31 @@ def set_zaak(group):
...
+            description=_("The name of the group assigned to the activity."),
+            location=OpenApiParameter.QUERY,
+        )
+    ],
+)
+class WorkStackGroupAdhocActivitiesView(WorkStackAdhocActivitiesView):
+    def get_activities(self) -> List[dict]:
+        group_assignee = self.request.query_params.get("group_assignee")
+        if not group_assignee:
+            return []
```
annashamray, line 97: Have you discussed with FE that they will have to know the name of the particular group?

I thought that the werkvooraad will show all activities grouped by the user groups which this user is the part of. 
For now if FE wants to do it they have to request usergroups and then to request activities for each user group. 

Could you make sure that this is the expected and desired?

backend/src/zac/activities/api/serializers.py
```diff
@@ -22,7 +23,12 @@ class Meta:
 
 
 class ActivitySerializer(serializers.ModelSerializer):
-    assignee = serializers.SlugRelatedField(
+    group_assignee = serializers.SlugRelatedField(
```
annashamray, line 26: some help-text is appreciated. For now it's not clear is it `id` or `name` of the group

backend/src/zac/activities/migrations/0003_auto_20211008_1136.py
```diff
@@ -0,0 +1,45 @@
...
+            field=models.ForeignKey(
+                blank=True,
+                help_text="Group responsible for managing this activity.",
+                null=True,
+                on_delete=django.db.models.deletion.SET_NULL,
+                to="auth.Group",
+                verbose_name="group assignee",
+            ),
+        ),
+        migrations.AddField(
```
annashamray, line 33: I think it should me `RenameField` to save all the data

annashamray (commented):
Overall it looks very nice

PR #479, opened 2021-11-02
Swap to unprivileged nginx container for app container

sergei-maertens:
This ensures no containers run as root.


sergei-maertens:
Verified locally

PR #480, opened 2021-11-02
General project upgrades/improvements

sergei-maertens:
* performance logging disabled by default
* swap out deprecated raven for sentry-sdk

### DIFF
--- a/backend/src/zac/conf/docker.py
+++ b/backend/src/zac/conf/docker.py
@@ -9,9 +9,3 @@ os.environ.setdefault("DB_CONN_MAX_AGE", "60")
 os.environ.setdefault("LOG_STDOUT", "yes")
 
 from .production import *  # noqa isort:skip
-
-#
-# Custom settings
-#
-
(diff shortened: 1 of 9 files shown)


backend/src/zac/conf/docker.py
```diff
@@ -9,9 +9,3 @@
 os.environ.setdefault("LOG_STDOUT", "yes")
 
 from .production import *  # noqa isort:skip
-
-#
-# Custom settings
-#
-
-ENVIRONMENT = "docker"
```
damm89, line 17: We're removing docker environment? Is unused?

damm89 (approved):
:+1:

backend/src/zac/conf/docker.py
sergei-maertens: nope, it's now fetched from an environment variable in the `base.py` settings. We only deploy using containers and the `ENVIRONMENT` is specified in the infrastructure code.

PR #481, opened 2021-11-03
Feature/contezza documents

wasyllopatka:
contezza-documentlist web-element updated.


frontend/zac-ui/libs/features/contezza-document-search/src/app.config.json
```diff
@@ -0,0 +1,15 @@
...
+  "pagination": {
+    "size": 25,
+    "supportedPageSizes": [5, 10, 25, 50, 100, 150]
+  },
+  "application": {
+    "name": "Contezza ZAC Documentlist"
+  },
+  "SEARCH_QUERY": "TYPE:'cm:content'",
+  "SEARCH_FIELDS": ["cm:name", "cm:title", "cm:description"],
+  "ALFRESCO_PREVIEW_URL": "https://alfresco.cg-intern.ont.utrecht.nl"
```
svenvandescheur, line 14: How do we deal with different environments like production?

frontend/zac-ui/proxy.conf.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
+  '/alfresco': {
+    target: 'https://alfresco-tezza.aks.utrechtproeftuin.nl',
```
svenvandescheur, line 3: How do we deal with different environments?

ingress/default.conf.template
```diff
@@ -27,4 +27,9 @@ server {
     location = /50x.html {
         root   /usr/share/nginx/html;
     }
+
+    location /alfresco/ {
+        proxy_pass https://alfresco.cg-intern.ont.utrecht.nl;
```
svenvandescheur, line 32: How do we deal with different environments?

svenvandescheur (commented):
Looks good to me, I would like to know how we would deal with various environments and preferably have a single source of truth for these kind of settings.

frontend/zac-ui/libs/features/contezza-document-search/src/app.config.json
wasyllopatka: Hi Sven, thank u. Values in app.config.json may by overriden on start docker container with variables.

frontend/zac-ui/proxy.conf.js
wasyllopatka: 'proxy.conf.json' is used only in development mode. So, this is not necessary.

ingress/default.conf.template
wasyllopatka: Here we can use variable instead of hardcoded url, e.g. 'proxy_pass ALFRESCO_URL;'

svenvandescheur: Is that currently already possible or de we still need to implement this?

frontend/zac-ui/libs/features/contezza-document-search/src/app.config.json
svenvandescheur: Is that currently already possible or de we still need to implement this?

wasyllopatka: We still need to implement this.

ingress/default.conf.template
wasyllopatka: I have this already implemented.

frontend/zac-ui/libs/features/contezza-document-search/src/app.config.json
sergei-maertens: This is crucial for when we deploy the app to ACC environment too (which should happen Soon :tm: )

frontend/zac-ui/libs/features/zaak-detail/src/lib/documenten/documenten.service.ts
```diff
@@ -26,6 +27,24 @@ export class DocumentenService {
...
+    const endpoint = encodeURI("/api/core/vertrouwelijkheidsaanduidingen");
+    return this.http.Get<any>(endpoint);
+  }
+
+  postDocument(formData: FormData): Observable<Document> {
+    return this.http.Post<any>(encodeURI(`/api/core/cases/document`), formData);
+  }
+
+  patchDocument(formData: FormData, bronorganisatie: string, identificatie: string): Observable<any> {
+    return this.http.Patch<any>(encodeURI(`/api/core/cases/${bronorganisatie}/${identificatie}/document`), formData);
```
sergei-maertens, line 40: This endpoint no longer takes `bronorganisatie` or `identificatie` parameters, see https://zac.cg-intern.ont.utrecht.nl/api/docs/#operation/core_cases_document_partial_update

Instead, the `zaak` URL in the request body is required.

ingress/Dockerfile
```diff
@@ -4,7 +4,9 @@ ENV BACKEND_HOST=backend \
     BACKEND_PORT=8000  \
     FRONTEND_HOST=frontend  \
     FRONTEND_PORT=8080 \
-    FRONTEND_PATH=/
+    FRONTEND_PATH=/ \
+    ACS_PROXY_URL=https://alfresco.cg-intern.ont.utrecht.nl
```
sergei-maertens, line 8: Is this (proxy) needed for production deployment as well or only for development?

The ingress setup in this repo is not used in the Kubernetes deployments - this is only for local full-stack environments and I'm trying to estimate the impact.

sergei-maertens (changes requested):
One endpoint bug spotted, one question :)

sergei-maertens:
@wasyllopatka it's not clear from the code to me, but how does the end-user in the ZAC now authenticate to Alfresco?

frontend/zac-ui/libs/features/zaak-detail/src/lib/documenten/documenten.service.ts
wasyllopatka: We do not use this endpoint for 'contezza-documentlist'.
But yes, it also has to be adjusted in the app.

wasyllopatka:
> @wasyllopatka it's not clear from the code to me, but how does the end-user in the ZAC now authenticate to Alfresco?

If user is not authenticated, he will get dialog to login to Alfresco.

frontend/zac-ui/libs/features/contezza-document-search/src/app.config.json
sergei-maertens: I am coordinating with Tahir about this - we might solve this at the infrastructure level so that `$zacDomain/alfresco/` is provided.

We don't deploy dynamic nginx images (though the entrypoint with find/replace and/or envsubst is a possible fallback solution)

frontend/zac-ui/libs/features/zaak-detail/src/lib/documenten/documenten.service.ts
sergei-maertens: I'm asking Sven to follow up on this!

sergei-maertens (approved):
I think this is okay for our needs? I have no idea if this is still blocked for merge or not though.

PR #482, opened 2021-11-03
Feature/group management

damm89:
Fixes the backend for: https://github.com/GemeenteUtrecht/ZGW/issues/935


backend/src/zac/accounts/api/serializers.py
```diff
@@ -65,16 +65,48 @@ class GroupSerializer(serializers.ModelSerializer):
...
+        return group
+
+    @transaction.atomic()
+    def update(self, instance, validated_data):
+        users = validated_data.pop("user_set")
+        old_users = instance.user_set.all()
+        remove_users = [old_user for old_user in old_users if old_user not in users]
+        add_users = [new_user for new_user in users if new_user not in old_users]
+        instance.user_set.add(*add_users)
+        instance.user_set.remove(*remove_users)
```
sergei-maertens, line 106: can't you use `instance.user_set.set(expected_users)` instead? which does the adding/removing under the hood

sergei-maertens (approved):
Minor remark that I'm not even 100% sure about, otherwise looks good

backend/src/zac/accounts/api/serializers.py
damm89: Yes, you're right.

PR #485, opened 2021-11-08
Include version/build information in image

sergei-maertens:
* Added baking in the build information to the container image
* Added middleware to emit the build information as response headers


annashamray (approved):
It would be nice to include these response headers into API schema

sergei-maertens:
> It would be nice to include these response headers into API schema

created #486 for that!

damm89 (approved):
Little bit confusing with the different variable names but seems to work.

PR #492, opened 2021-11-10
Feature/bpmn xml endpoint

damm89:
Backend for https://github.com/GemeenteUtrecht/ZGW/issues/914


backend/src/zac/camunda/api/views.py
```diff
@@ -337,3 +337,26 @@ def post(self, request, *args, **kwargs):
...
+    )
+    def get(self, request, *args, **kwargs):
+        client = get_client()
+        process_definition_id = kwargs["process_definition_id"]
+        response = client.get(f"process-definition/{process_definition_id}/xml")
+        serializer = self.serializer_class(data=response)
+        try:
+            serializer.is_valid(raise_exception=True)
+        except exceptions.ValidationError:
+            return Response(response, status=status.HTTP_400_BAD_REQUEST)
```
annashamray, line 360: Hmm I think it's more logical to check the HTTP status of camunda request instead of serializer validation here. 
I suppose that serializer validation is used for input parameters, but here it's another situation -  we don't have a user input, we just want to make sure that the API call was successful, so I'd suggest using `response.raise_for_status()` here

annashamray (commented):
Could you please check the CI error? It's either a glitch or django-hijack-admin should be updated

backend/src/zac/camunda/api/views.py
damm89: The camunda client .get (and the called .request) method(s) don't return a response object, only the response content. :(

PR #500, opened 2021-11-19
:sparkles: Allow search on partial value of eigenschappen

damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/981

### DIFF
--- a/backend/src/zac/elasticsearch/drf_api/serializers.py
+++ b/backend/src/zac/elasticsearch/drf_api/serializers.py
@@ -1,3 +1,5 @@
+import re
+
 from django.utils.translation import gettext_lazy as _
 
 from rest_framework import serializers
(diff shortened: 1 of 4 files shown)


annashamray:
@damm89 some testting would be appreciated

damm89:
> @damm89 some testing would be appreciated

I was just checking to see if you think this would suffice. :-)

backend/src/zac/elasticsearch/drf_api/serializers.py
```diff
@@ -70,6 +72,9 @@ class SearchSerializer(serializers.Serializer):
         default=False,
     )
 
+    def validate_omschrijving(self, omschrijving):
+        return re.escape(omschrijving)
```
annashamray, line 76: Could you explain why to escape `omschrijving` ?

annashamray (commented):
Partial search works, but it's case sensitive. 
I suppose it's because zaak eigenschap value is saved as Keyword field, not as Text, which doesn't allow full-text search. 
But I think it can be done in the next step.

backend/src/zac/elasticsearch/drf_api/serializers.py
damm89: Because otherwise it won't find anything if omschrijving has special characters (such as a " ").

annashamray (approved):

PR #501, opened 2021-11-22
Issue/zgw#989 zgw#1011

damm89:
Relates to https://github.com/GemeenteUtrecht/ZGW/issues/989 and https://github.com/GemeenteUtrecht/ZGW/issues/1011.

### DIFF
--- a/backend/src/zac/core/api/serializers.py
+++ b/backend/src/zac/core/api/serializers.py
@@ -39,6 +39,7 @@ from zac.core.services import (
     fetch_zaaktype,
     get_document,
     get_documenten,
+    get_informatieobjecttypen_for_zaaktype,
     get_statustypen,
     get_zaak,
 )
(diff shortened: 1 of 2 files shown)


annashamray (changes requested):
Afaik all catalogi-related fields should be immutable in ZGW APIs, so if you try to change `zaak.zaaktype` or `document.informatieobjecttype` or `besluit.besluittype` you'll receive 400 error.
It's true for reference implementation of Documenten API, but I can't find the mention of it in API schema. Perhaps the OAS should be updated. 

@damm89 does changing `informatieobjecttype` work for you locally? If yes, what documenten API are you using? If it allows changing `informatieobjecttype` this should be considered as a bug now. 

@sergei-maertens correct me if I'm mistaken, please

damm89:
> Afaik all catalogi-related fields should be immutable in ZGW APIs, so if you try to change `zaak.zaaktype` or `document.informatieobjecttype` or `besluit.besluittype` you'll receive 400 error. It's true for reference implementation of Documenten API, but I can't find the mention of it in API schema. Perhaps the OAS should be updated.
> 
> @damm89 does changing `informatieobjecttype` work for you locally? If yes, what documenten API are you using? If it allows changing `informatieobjecttype` this should be considered as a bug now.
Yes it does actually... And not just locally. I'm able to change the document type of a document in a recent zaak locally and zac-test will show the changes. I'm using the drc.utrechtproeftuin.nl.

> @sergei-maertens correct me if I'm mistaken, please

So course of action in case the user story doesn't get changed would be to change the document.

sergei-maertens:
`drc.utrechtproeftuin.nl` is not standards compliant, and apparently this is another case where the validation fails. Checked the reference implementation too, and that one is bugged in the sense that it's allowed to change and there are also no tests, but this was _not_ intentional!

PR #503, opened 2021-11-23
As a zac user i want to be able to browse and search on alfresco

svenvandescheur:
Co-authored-by: Sven van de Scheur <svenvandescheur@gmail.com>
Co-authored-by: Kelvin Chan <kelvinchancy95@gmail.com>
Co-authored-by: wasyl <wasyl.lopatka@contezza.nl>
Co-authored-by: Sergei Maertens <sergei@maykinmedia.nl>


svenvandescheur:
Squashed from: https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/481

svenvandescheur:
I confirm that this is working with the changes made by Contezza. Currently de proxy runs through nginx. @annashamray and @damm89 stated that they would prefer the proxy to go through BFF as well. Is this something we want to fix before the merge or a a future improvement?

sergei-maertens:
Tahir and I set up the `zac.cg-intern.ont.utrecht.nl/alfresco/` endpoint - it was my understanding that the client code hits the `/alfresco/` URL and that we need to handle the proxying on the infrastructure level. This would absolutely have my preference since this is not code I want to maintain and building proxies is hard too.

It has its own auth mechanism as far as I understand and cannot be tied in with the Django user session.

svenvandescheur:
@sergei-maertens Can you discuss this with the other backenders? I'm have no preference, your correct w/r to the proxy.

sergei-maertens (approved):
Disclaimer: I reviewed the original PR, 100+ files really is not manageable to properly review

annashamray (approved):
Approve the approach with integrating with Alfresco lib in frontend without proxying through BFF

Issue #505, opened 2021-11-24
Update fetch_process_instances


sergei-maertens:
With many nested subprocesses, we're seeing long request times because all subprocesses are probed sequentially. We can optimize this by processing sub-processes in parallel and speeding up the BFF endpoint in general.

PR #510, opened 2021-11-29
Issue/#505

damm89:
Related to #505 .

Response time is roughly decreased by ~40-80% (more so if there are a lot of subprocesses).
For zaak 1697 in zac-test.utrechtproeftuin.nl which has 10 sub processes the time is reduced from ~5.5-6 seconds to ~1.3-1.7 seconds.
For zaak 1705 in zac-test.utrechtproeftuin.nl which has 2 sub processes the time is reduced from ~1 seconds to ~0.6 seconds.

https://github.com/GemeenteUtrecht/zaakafhandelcomponent/pull/511 is a small improvement on top of this, but I think less desirable.


sergei-maertens (approved):
Nice timing results!

PR #512, opened 2021-12-01
Feature/manage groups

damm89:
Superusers should be allowed to manage and view all groups and admin users should be allowed to determine who gets to manage which group(s).


annashamray (approved):
:+1:

PR #526, opened 2021-12-09
Fix/sentry 7288e3650d0045c68a84e24ce5200a09

damm89:
Fixes https://sentry.maykinmedia.nl/maykin-media/zac-utrecht-acc/issues/298428/events/3874302/.

### DIFF
--- a/backend/src/zac/camunda/api/views.py
+++ b/backend/src/zac/camunda/api/views.py
@@ -66,6 +66,7 @@ class ProcessInstanceFetchView(APIView):
         zaak_url = request.GET.get("zaak_url")
         if not zaak_url:
             err_serializer = ErrorSerializer(data={"detail": "missing zaak_url"})
+            err_serializer.is_valid()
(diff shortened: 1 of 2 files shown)


backend/src/zac/camunda/api/views.py
```diff
@@ -66,6 +66,7 @@ def get(self, request: Request, *args, **kwargs):
         zaak_url = request.GET.get("zaak_url")
         if not zaak_url:
             err_serializer = ErrorSerializer(data={"detail": "missing zaak_url"})
+            err_serializer.is_valid()
```
annashamray, line 69:
```suggestion
            err_serializer = ErrorSerializer({"detail": "missing zaak_url"})
```

I don't quite remember why the serializer is needed here, since the data is always the same. But I think that validating the known data is a bit excessive, but I'll live it up to you.

annashamray (approved):

PR #544, opened 2021-12-31
:bug: Add get_or_create to activityfactory to prevent db integrity error

damm89:
See https://github.com/GemeenteUtrecht/zaakafhandelcomponent/runs/4667886103?check_suite_focus=true


annashamray (approved):
:+1: TIL!

PR #548, opened 2022-01-06
:boom: Add group to review data from the assignee of the camunda user …

damm89:
…task.

Requires frontend changes: @kelvincy

Fixes https://github.com/GemeenteUtrecht/ZGW/issues/1132

Relies on https://github.com/GemeenteUtrecht/ZGW/pull/1141

### DIFF
--- a/backend/src/zac/camunda/data.py
+++ b/backend/src/zac/camunda/data.py
@@ -49,7 +49,7 @@ class Task(_Task):
     def has_form(self) -> bool:
         return bool(self.form)
 
-    def get_variable(self, name: str, default: Optional[str] = None) -> Any:
+    def get_variable(self, name: str, default: Optional[Any] = None) -> Any:
         return get_task_variable(self.id, name, default=default)
 
     def assignee_type(self) -> str:
--- a/backend/src/zac/camunda/data.py
+++ b/backend/src/zac/camunda/data.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass, field
-from typing import Any
+from typing import Any, Optional
 from uuid import UUID
 
 from django.contrib.auth.models import Group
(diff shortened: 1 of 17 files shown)


backend/src/zac/contrib/kownsl/views.py
```diff
@@ -124,33 +126,63 @@ def get(self, request, request_uuid):
         return Response(serializer.data)
 
     def post(self, request, request_uuid):
+        if not (task_id := request.query_params.get("taskid")):
+            raise exceptions.ValidationError("'taskid' query parameter is required.")
+
+        task = get_task(task_id)
+        data = {**request.data}
+        if task.assignee_type() == AssigneeTypeChoices.group:
+            data["group"] = f"{task.assignee}"
```
annashamray, line 135: I don't quire understand the logic behind. Here the advice or the approval is created, right? 
But the user can create the review for themselves not for the group. I thought `group` attribute should come from front-end. And in this case this won't be a breaking change

damm89: Talked it over in Slack, hopefully enough to understand why Im doing this here. Otherwise please see https://github.com/GemeenteUtrecht/ZGW/issues/1132 for more context.

backend/src/zac/camunda/data.py
```diff
@@ -49,8 +49,8 @@ class Task(_Task):
     def has_form(self) -> bool:
         return bool(self.form)
 
-    def get_variable(self, name: str) -> Any:
-        return get_task_variable(self.id, name)
+    def get_variable(self, name: str, default: Optional[str] = None) -> Any:
```
damm89, line 52:
```suggestion
    def get_variable(self, name: str, default: Optional[str] = Any) -> Any:
```

PR #556, opened 2022-01-25
Bump/drf spectacular

damm89:


backend/src/zac/conf/includes/base.py
```diff
@@ -458,6 +458,10 @@
     "EXTERNAL_DOCS": {
         "url": "https://zaakafhandelcomponent.readthedocs.io/",
     },
+    "ENUM_NAME_OVERRIDES": {
```
annashamray, line 461: :+1: TIL

backend/src/zac/core/api/views.py
```diff
@@ -734,11 +736,13 @@ class VertrouwelijkheidsAanduidingenView(ListAPIView):
...
     serializer_class = VertrouwelijkheidsAanduidingSerializer
 
-    def get_queryset(self):
-        return [
+    def get(self, request):
+        instances = [
             VertrouwelijkheidsAanduidingData(label=choice[1], value=choice[0])
             for choice in VertrouwelijkheidsAanduidingen.choices
         ]
+        serializer = self.serializer_class(instance=instances, many=True)
```
annashamray, line 744: Now drf spectacular ignores `many=True` and generated schema looks like JSON object not the array of the objects

Options to avoid it:
* use `@extend_schema(responses=... )`
* drf-spectacular looks for `get_serializer` method, so it can be added
* use `zac.api.mixins.ListMixin`, which was created for this purpose

backend/src/zac/werkvoorraad/api/views.py
```diff
@@ -94,22 +95,23 @@ def get_activities(self) -> List[dict]:
...
         zaken = search(
-            user=self.request.user,
-            behandelaar=self.request.user.username,
+            user=request.user,
+            behandelaar=request.user.username,
             ordering=ordering,
         )
         unfinished_zaken = [zaak for zaak in zaken if not zaak.einddatum]
-        return unfinished_zaken
+        serializer = self.serializer_class(instance=unfinished_zaken, many=True)
```
annashamray, line 113: same here - the schema doesn't have an array in the response

backend/src/zac/core/api/views.py
damm89: Thank you!!

annashamray (approved):

PR #557, opened 2022-01-25
#1155 Zaakinformatievelden

damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/1155

### DIFF
--- a/backend/src/zac/core/api/serializers.py
+++ b/backend/src/zac/core/api/serializers.py
@@ -39,6 +39,7 @@ from zac.core.services import (
     fetch_zaaktype,
     get_document,
     get_documenten,
+    get_informatieobjecttypen_for_zaak,
     get_statustypen,
     get_zaak,
 )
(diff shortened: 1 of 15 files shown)


backend/src/zac/core/api/serializers.py
```diff
@@ -599,6 +599,20 @@ class Meta:
         extra_kwargs = {"url": {"read_only": True}}
 
 
+class CreateZaakEigenschapSerializer(serializers.Serializer):
+    naam = serializers.CharField(
+        help_text=_(
+            "Name of EIGENSCHAP. Must match EIGENSCHAP name as defined in Catalogi API."
+        )
+    )
+    waarde = serializers.CharField(
```
annashamray, line 608: The field names are not consistent for request and response, and they are also different for PATCH request and POST request
I'd suggest renaming to avoid confusion, but if FE is ok with these field names, it's not required

annashamray (commented):
I have only one remark:

PR #563, opened 2022-01-28
:bug: Set permissions choices on serializer field correctly

damm89:
Fixes https://github.com/GemeenteUtrecht/ZGW/issues/1148


annashamray (approved):
* I think there should be a data migration added with renaming `activities_read` to `activiteiten_inzien` for existing entries in `AtomicPermission` and `Role` models

* Could you also rename it in the documentation (`doc/autjorization.rst`) ?

PR #564, opened 2022-01-31
:sparkles: Only show user credential login screen on non-production env

damm89:
Fixes https://github.com/GemeenteUtrecht/ZGW/issues/1192


annashamray (approved):
For future I'd suggest adding a boolean to the `CoreConfig` model which will allow or forbid using non-ADFS authentication

PR #565, opened 2022-02-02
:boom: Return validation error on get review if review has alread…

damm89:
…y been submitted by assignee.

Fixes https://github.com/GemeenteUtrecht/ZGW/issues/1180.

Requires changes to frontend: on 
http://localhost:8080/api/docs/#operation/kownsl_review_requests_advice_retrieve, and
http://localhost:8000/api/docs/#operation/kownsl_review_requests_approval_retrieve, 
please include the `assignee` query parameter.


svenvandescheur:
I can't see the assignee query parameter in the doc as linked (in this branch), is that correct?

PR #639, opened 2022-04-19
:lipstick: #1402 - UX: (Related objects block) - As a user I want to …

svenvandescheur:
…be able to search large amount of objects when adding a related object.


kelvincy:
@svenvandescheur I added some styling to make it look cleaner

kelvincy (approved):

PR #643, opened 2022-04-23
Feature/fe redesign

kelvincy:
Redesign for:
https://github.com/GemeenteUtrecht/ZGW/issues/1393
https://github.com/GemeenteUtrecht/ZGW/issues/1394
https://github.com/GemeenteUtrecht/ZGW/issues/1403
https://github.com/GemeenteUtrecht/ZGW/issues/1405


svenvandescheur (changes requested):
Few remarks:

- Can you have a look at relative imports in the scss file?
- Can you put the map in a card?
- There seems to be an issue with spamming messages if loading tasks does not succeed:

Apr-25-2022 11-04-05

PR #683, opened 2022-06-21
Optimise huisstijl styling

kelvincy:


frontend/zac-ui/libs/shared/ui/components/src/lib/elements/message/message.component.scss
```diff
@@ -6,12 +6,12 @@
...
   &--primary {
-    border-color: darken($color-info, 20%);
-    background-color: $color-info;
+    border-color: darken($color-tertiary, 20%);
+    background-color: $color-tertiary;
   }
   &--warn {
-    border-color: darken($color-warning, 20%);
-    background-color: $color-warning;
+    border-color: darken($color-background-lightyellow, 20%);
```
svenvandescheur, line 13: Can we find a different solution to this than seemingly randomly mapping colors? I'm more in favor of having meaningful references to colors (like primary, secondary, info or warn) and then using some kind of config to map that out to the actual colors.

Is there a reason why we divert from this here?

frontend/zac-ui/libs/shared/ui/components/src/lib/elements/progress-bar/progress-bar.component.scss
```diff
@@ -4,17 +4,17 @@
...
     height: 2.25rem;
-    border: 1px solid $color-border;
-    border-radius: 5px;
+    border-radius: 10px;
   }
+  border-radius: 5px;
   width: 0;
   height: 100%;
-  background-color: $color-info;
+  background-color: $color-tertiary;
```
svenvandescheur, line 12: Same, see remark above.

svenvandescheur (changes requested):

PR #729, opened 2022-08-16
:bug: Fix format eigenschappen injected into camunda process

damm89:


damm89:
Ran local tests - merge.

PR #786, opened 2022-11-02
:bug: Use identificatie instead of omschrijving to create a zaak

damm89:


damm89:
Fixes a bug related to zaak creation based on `omschrijving` field. This field should be the `identificatie` field.

PR #788, opened 2022-11-03
:sparkles: #1824 - Filteren op objecttyperegistraties UI-ZAC default …

svenvandescheur:
…gefilterd + uitwijkoptie.


damm89 (commented):
Works functionally - placement of checkbox feels a bit user unfriendly.

PR #1011, opened 2023-07-05
:sparkles: Allow the possibility of user tasks to not be finished by …

damm89:
…a user


damm89:
Related to https://github.com/GemeenteUtrecht/ZGW/issues/2146

PR #1022, opened 2023-07-17
:sparkles: Return full name when a checklist is locked and returns an…

damm89:
… error


damm89:
Resolves https://github.com/GemeenteUtrecht/ZGW/issues/2155

Issue #499, opened 2021-11-16
Upgrade to django-hijack 3


sergei-maertens:
* hijack 2.x and hijack-admin currently cannot install with setuptools > 58 (see https://setuptools.pypa.io/en/latest/history.html#v59-0-0)
* hijack-admin is unmaintained and no longer needed for django-hijack 3
* django-hijack 3 is a rewrite, best is to remove integration completely and re-do the setup

PR #1028, opened 2023-07-21
Bump/hijack and cors

damm89:


damm89:
Fixes https://github.com/GemeenteUtrecht/zaakafhandelcomponent/issues/499

PR #1247, opened 2024-04-15
:bug: Return validation error if omschrijving is longer than 80 chara…

damm89:
…cters


damm89:
Fixes https://github.com/GemeenteUtrecht/ZGW/issues/2403

PR #1286, opened 2024-12-06
:ok_hand: Process feedback regarding recently viewed section on landi…

damm89:
…ng page


damm89:
Fixes https://github.com/GemeenteUtrecht/ZGW/issues/2506

